PR #2, opened 2016-04-21
Split all edition level XML files out and placed them in the correct …

sonofmun:
…directories. Also produced a CSV for all metadata for these editions.


sonofmun:
Merging my own PR.

PR #8, opened 2016-05-04
The repository should now be able to be tested with Hook

sonofmun:


sonofmun:
@annettegessner Don't forget to delete the old branch!

PonteIneptique:
Do not know if normal but I found a number of texts having only one citation node :')

sonofmun:
@PonteIneptique That was the case on several of the texts. They were fragments or a block of text from someone that had no citation scheme whatsoever. I added a single node for all of these texts and we can decide later if they need something deeper or not.

PonteIneptique:
Cool :)

PonteIneptique:
(Note : you should add the badges to brag about it ;) )

annettegessner:
In PL there are a number of works only having one level of citation, so this could be correct, but we'll have to check it.

sonofmun:
@PonteIneptique I was just about to do that but was waiting for the PR to be merged.

PR #9, opened 2016-05-04
Added Hook Badges to README

sonofmun:


sonofmun:
Hm, I guess I should have just committed to master for the change to the README. Oh well, too late now.

Issue #68, opened 2016-05-04
Fix /tlg4193/tlg012/tlg4193.tlg012.opp-grc1.xml


sonofmun:
error: attribute "type" not allowed here [In (L69 C54)]

Issue #46, opened 2016-05-04
Fix /tlg2768/tlg002/__cts__.xml


sonofmun:
URNs testing failed
Group urn : urn:cts:greekLit:tlg2768
Work urn : urn:cts:greekLit:tlg2768.tlg002.opp
Editions and translations urns : urn:cts:greekLit:tlg2768.tlg002.opp.grc1

PR #69, opened 2016-05-04
Corrected Issue #68

sonofmun:


sonofmun:
This is to correct an error in the URN and the naming convention mentioned issue #46 The name of the branch is actually incorrect.

sonofmun:
File is now correct. PR can be merged.

PR #74, opened 2016-05-11
issue68

annettegessner:
corrected TEI header, added chapters according to numbers in text (ignored section numbers on the side of pages since the start and end of those were hard to determine)


annettegessner:
passing hook test now - delete issue #68 and branch issue68 when merging, please

PR #172, opened 2016-05-19
idno to ref target

annettegessner:
changed all headers to contain
`<ref target=".*</ref>`
instead of
`<idno n=".*</idno>`


sonofmun:
Didn't break anything, so I guess I will merge.

Issue #168, opened 2016-05-19
Ancient Greek, not modern - correct all language ident


annettegessner:
In many files you find something like this in the header

`<langUsage>
<language ident="el">Greek</language>
</langUsage>`

since our files have ancient Greek, all the instances of 
`<language ident="el">Greek</language>` (i.e. modern Greek)
have to be be changed to
`<language ident="grc">Greek</language>` (i.e. ancient Greek)

annettegessner:
PR #169

annettegessner:
Addition:
correct
<language ident="greek">Greek</language>
into
<language ident="grc">Greek</language>

PR #177, opened 2016-05-24
issue 168 second correction on language tags

annettegessner:
corrected
`<language ident="greek">Greek</language>`
into correct Epidoc
`<language ident="grc">Greek</language>`


annettegessner:
didn't break (or improve) anything in the hook test, as it should,
close #168 when merging

Issue #174, opened 2016-05-19
fix text tags


annettegessner:
12 files do not just have a pure text-tag, but some additional information in this tag like
`<text xml:lang="grc">`
or 
`<text xml:id="tlg0555.tlg004b.opp-grc1">`

I guess, we should normalize this.

PR #179, opened 2016-05-24
issue174 normalized text tags

annettegessner:
normalized all text-tags and made sure, information included in an attribute was preserved elsewhere, issue #174


annettegessner:
didn't break (or improve) anything in the hook test, as it should,
close #174 when merging

Issue #152, opened 2016-05-12
Only 1 citation unit in /tlg4024/tlg002/tlg4024.tlg002.opp-grc12.xml


sonofmun:

annettegessner:
look for scan!

Issue #40, opened 2016-05-04
Fix /tlg2200/tlg00544/tlg2200.tlg00544.opp-grc1.xml


sonofmun:
Duplicate references found : 28

Issue #39, opened 2016-05-04
Fix /tlg2200/tlg00543/tlg2200.tlg00543.opp-grc1.xml


sonofmun:
Duplicate references found : 3, 5

PR #181, opened 2016-05-25
la to lat and 3 issues

annettegessner:
changed all "la" to "lat" (concerning language) and worked on 3 issues:
#39 
#40 
#152


annettegessner:
all 3 files passing hook test

Issue #41, opened 2016-05-04
Fix /tlg2200/tlg00545/tlg2200.tlg00545.opp-grc1.xml


sonofmun:
Duplicate references found : 6, 2, 7

annettegessner:
chapters wrong, needs to be reorganized entirely

PR #184, opened 2016-05-26
Tabea issue #41

planatheisa:
corrected structure


sonofmun:
@planatheisa Whatever you did broke Capitains compliance of the file and now it can't even be read by Capitains.

planatheisa:
Well, I have no idea how. All I did was change the div structure, add myself in responsibility statement and change document language to greek. Also changed CTS scheme to fit the new div tags.

sonofmun:
Don't worry about changing the responsibility statement. I will be doing a wholesale change to the TEI headers to include everyone who works on the repository.  Try taking that change out and see what happens.

planatheisa:
I took the original file from the split file because I didn't know it was in data but I just compared both and the only difference, except those mentioned just now, was that my document still had "la" instead of "lat" for Latin.

sonofmun:
Try again using the one from data instead. You should only be using files from the data folder for everything you are doing on this repo.

planatheisa:
OK. I'll redo it from scratch and see what happens.

Issue #162, opened 2016-05-12
Only 1 citation unit in /tlg9004/tlg001/tlg9004.tlg001.opp-grc1.xml


sonofmun:

annettegessner:
this is a commentary on another work
https://archive.org/stream/commentariaina13pt3akaduoft#page/570/mode/2up

there is always one paragraph `<p>` citing the original (e.g.: p.94b18) and one paragraph of commentary after this

suggestion: either number every paragraph or every two paragraphs, either chronologically (1, 2, 3, ...) or according to the original (e.g.: p.94b18)

sonofmun:
I would suggest chronologically instead of having the numbering tied to a single edition. Unless, of course, this is the definitive edition that everyone bases their citation scheme on.
Also, why not number them 1a (for the original) and 1b (for the commentary) or some such so that even in the citation scheme we can tell which parts are related to each other? Or is that too complex to integrate into the existing XML?

annettegessner:
numbering chronologically should be easy, but it seems we have a lot of comments and I am unsure about the Aristotelian citations in this text being "the definite edition", but we definitely encounter this problem, when commenting on the Bible, where the citation used in the text should be definitive, see #161

annettegessner:
so the question for all commentaries is: should we (try to) keep the citations used in the text as part of our CTS scheme or not

annettegessner:
I suggest numbering consecutively for now.

@planatheisa Could you take every instance of 
`<p>p.` (followed by numbers and letters like "891b23" or "90 a 5")
and add a numbered div in front  of it 
`<div type="textpart" subtype="chapter" n="1">`
(just get rid of the first and only chapter-div-tag named n="3").
Thanks!

annettegessner:
Mark as type="commentary"?!

PR #194, opened 2016-05-26
issue #162

planatheisa:
added numbered div-tags, also added ref-tag


annettegessner:
The file doesn't pass the hook test:

Duplicate references found : 10, 44

planatheisa:
Now it passes.

Issue #161, opened 2016-05-12
Only 1 citation unit in /tlg4102/tlgX03/tlg4102.tlgX03.opp-grc1.xml


sonofmun:

annettegessner:
this is a commentary on New Testament book Revelation. The chapter and line of the Bible, which are first cited and then commented, can be found at the left side of the scanned page) hard to determine, since the XML doesn't divide numbering from right or left side (line numbering according to book (5, 10, 15, ...) , i.e. not relevant for citing) of the scanned page) and on top of the scanned page

https://archive.org/stream/catenaegraecorum08cramuoft#page/496/mode/2up

every paragraph is either a line from the Bible or the commentary on it, but not every paragraph containing original text has it's own citation, since it can be from the same line already cited

suggestion: maybe try marking every paragraph beginning with a numbered line break, e.g.:
`<lb n="9"/> <p>`
chapter has to be determined manually

PR #195, opened 2016-05-26
issue #161

annettegessner:


annettegessner:
passes Hook test, 
close issue #161, when merging

Issue #90, opened 2016-05-12
Edition information for tlg1205.tlg002.perseus-grc2.xml


sonofmun:
There is no edition information in the TEI header here. We need to find this.

annettegessner:
Hey @AlisonBabeu,
Do you by chance have the metadata to this URN 
tlg1205.tlg002.perseus-grc2 ([LINK](https://github.com/OpenGreekAndLatin/First1KGreek/blob/0cf3d91180ebb5edd7c6679a8a8ace050f79ed77/data/tlg1205/tlg002/tlg1205.tlg002.perseus-grc2.xml))
somewhere? I didn't find it in the Perseus Catalog, but since it's a Perseus URN, I hoped you might have it in some list.

We only have
`<titleStmt>
        <title>De resurrectione</title>
        <author>Athenagoras</author>
        <editor role="editor">cureditor</editor>
      </titleStmt>`
and that's it - no editor, publisher or date.

annettegessner:
Aha, found some hidden info, that will help identify this thing:
`<bibl n="Owen 1876 New York"/>`

AlisonBabeu:
So I've figured out the edition for this, largely because of the starting page number of the work, pg 53:
https://books.google.com/books?id=rh6x5Yx45t4C&pg=PA53#v=onepage&q&f=false
So it turns out the edition used for this is Athenagoras, edited by Francis Andrew March, published by Harper Brothers in 1876.
Full bibliographic info can be found here:http://www.worldcat.org/title/athenagoras/oclc/17997690.
And it turns out I do have a spreadsheet that may help you and Matt with tracking some of this down. I'll send a separate email with a link to that shortly.

PR #212, opened 2016-05-31
added edition info

annettegessner:
#90


annettegessner:
still passing hook test

Issue #45, opened 2016-05-04
Fix /tlg2703/tlg001/tlg2703.tlg001.opp-grclat1.xml TRANSLATION


sonofmun:
Duplicate references found : praef.1, praef.3, praef.4, 1.10, 1.8, 1.12, 1.1, 1.2, 1.9, 1.3, 1.5, 1.11, 1.16, 1.13, 1.6, 1.7, 1.4, 1.15, 1.14, 2.2, 2.5, 2.1, 2.7, 2.11, 2.4, 2.8, 2.9, 2.6, 2.10, 2.12, 2.3, 3.4, 3.6, 3.9, 3.10, 3.7, 3.3, 3.5, 3.8, 3.2, 3.11, 3.1, 3.12, 4.6, 4.1, 4.7, 4.8, 4.4, 4.2, 4.5, 4.3, 5.5, 5.3, 5.9, 5.4, 5.7, 5.6, 5.8, 5.1, 5.2, 6.6, 6.11, 6.9, 6.5, 6.1, 6.3, 6.12, 6.2, 6.14, 6.4, 6.13, 6.10, 6.7, 6.8, 7.6, 7.9, 7.4, 7.11, 7.5, 7.10, 7.8, 7.1, 7.7, 7.3, 8.3, 8.9, 8.1, 8.4, 8.2, 8.7, 8.8, 8.6, 9.4, 9.7, 9.6, 9.8, 9.3, 9.2, 9.10, 9.1, 9.9, 9.5, 10.2, 10.10, 10.5, 10.3, 10.7, 10.6, 10.11, 10.4, 10.1, 10.8, 10.9, 11.5, 11.10, 11.6, 11.7, 11.3, 11.4, 11.11, 11.2, 11.9, 11.1, 11.8, 11.12, 12.4, 12.6, 12.5, 12.7, 12.8, 12.2, 12.1, 12.3, 12.9, 13.11, 13.1, 13.9, 13.10, 13.12, 13.3, 13.8, 13.7, 13.4, 13.6, 13.5, 13.2, 14.3, 14.1, 14.4, 14.7, 14.8, 14.9, 14.5, 14.2, 14.6, 15.9, 15.5, 15.10, 15.2, 15.6, 15.8, 15.7, 15.4, 15.11, 15.3, 15.1

sonofmun:
[...]

PR #213, opened 2016-05-31
header corrections

annettegessner:
in
tlg2703.tlg001.opp-grc1
and
tlg2703.tlg001.opp-grclat1


annettegessner:
as before: first file still passes hook test, second file still doesn't
for further information about difficult second file, see (but don't close yet) #45

sonofmun:
If this does not improve our Hook results, is there a reason to merge this PR?

sonofmun:
Never mind, I see that you corrected the headers. I will merge it.

Issue #215, opened 2016-05-31
tlg2768.tlg002.opp-grc1


annettegessner:
Greek text of Gelasius with Latin translation under the footnotes.
Latin text sometimes in a paragraph, sometimes in a footnote. Div-tags for Latin text have to be added.
Since this work has only 9 physical pages, the file can be corrected manually. Otherwise it's problems are similar to the files mentioned in #211 .

annettegessner:
I split the files manually to see, if there were any other problems regarding this issue, but there weren't. The only thing that had to be looked up closely was the chapters for the Latin textparts, the rest could maybe even have been automatized.

annettegessner:
@sonofmun: Is there a "cool" way to find files including Latin translations? I found this by searching for files including the (randomly chosen) Latin word "quam" and it didn't have an issue assigned for it yet.

sonofmun:
I would think that looking for a common word would be the best way to go about it. Though these will also often appear in the Latin titles of Greek works, so it will also produce some noise.

PR #216, opened 2016-05-31
#215 split translation

annettegessner:
splitted Latin translation from Greek text manually, since it was only 9 pages and the div tags were better than in other files

### DIFF
--- a/data/tlg2768/tlg002/__cts__.xml
+++ b/data/tlg2768/tlg002/__cts__.xml
@@ -4,4 +4,8 @@
     <ti:label xml:lang="eng">According to Eusebius and Theognius</ti:label>
     <ti:description xml:lang="eng">Gelasius, According to Eusebius and Theognius</ti:description>
   </ti:edition>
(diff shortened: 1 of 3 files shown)


sonofmun:
/tlg2768/tlg002/tlg2768.tlg002.opp-lat1.xml needs to be added to the `__cts__.xml` file.

annettegessner:
forgot to change the __cts__ file.
Both files passing hook test now.

sonofmun:
Just add the following lines to `__cts__.xml` as a direct child of `<ti:work>`

    <ti:edition urn="urn:cts:greekLit:tlg2768.tlg002.opp-lat1" workUrn="urn:cts:greekLit:tlg2768.tlg002">
        <ti:label xml:lang="eng">According to Eusebius and Theognius</ti:label>
        <ti:description xml:lang="eng">Gelasius, According to Eusebius and Theognius</ti:description
    </ti:edition>

annettegessner:
geeeeez, I already figured all on my own and I made it "latinLit" instead of "greekLit"
:-P

sonofmun:
The Latin translation should still be in the greekLit domain, not in the latinLit domain. You need to change that back.

annettegessner:
done - passing hook test

Issue #210, opened 2016-05-31
How to handle drama?


annettegessner:
see #103 and #81

PonteIneptique:
Drama should be by line. Which does not mean that chores and other divs should not be in the XML ( We just simply do not use it in the final xpath `/TEI/TEXT/BODY/DIV//L[@n="$1"]` :) )

annettegessner:
Cool, so this means, we can leave the files as they are, as long as the lines are there and we just have to change the xPath, right? B-)

annettegessner:
So the refsDecl for all drama should look like this:
`<cRefPattern n="line" matchPattern="(.+)" replacementPattern="#xpath(/tei:TEI/tei:text/tei:body/tei:div[@type='edition']//tei:l[@n='$1'])"/>`
FYI @planatheisa

PonteIneptique:
And maybe you can drop @type="edition" ? :)

PonteIneptique:
By the way @annettegessner, do you know that

\`\`\`xml
`<a b="1">`
\`\`\`

actually formats for xml 

```xml
<a b="1">
```

annettegessner:
Ooops! I am trying this for those two files right now and if it works, I'll do that for ALL our files, opening a new issue.

annettegessner:
Niiice, thanks for the tip with ´´´xml !

jduff-chs:
This may be relevant I believe: there are a handful of instances in drama where two characters will speak on the same line. There are two in the Seven Against Thebes, for instance. Because we're nesting every line within a <sp> tag, this needs some workaround - I've chosen to subdivide into alphabetical divisions within the line, which Hook and Nemo are okay with. See the example below, where I split line 980 so it can have portions spoken by Antigone and Ismene. Is this a suitable practice to solve this problem?
![screen shot 2016-06-10 at 4 14 35 pm](https://cloud.githubusercontent.com/assets/19731635/15977591/b6031bba-2f27-11e6-84d2-2e82354a9965.png)
![screen shot 2016-06-10 at 4 14 49 pm](https://cloud.githubusercontent.com/assets/19731635/15977593/b879a2a6-2f27-11e6-8662-b33308efb6df.png)

annettegessner:
About duplicate line numbering:
In case of two lines wrongly having the same line number, add a linebreak, e.g.:
replace
```xml
<l n="236">δην, στόματός τε καλλιπρῴ-</l>
[...]

Issue #103, opened 2016-05-12
Check citation scheme and refsDecl for /tlg0085/tlg003/tlg0085.tlg003.opp-grc3.xml


sonofmun:

annettegessner:
drama with too many different subtypes in the div-tags and no n-element

Issue #81, opened 2016-05-12
tlg0085.tlg007.opp-grc3.xml milestones


sonofmun:
The sections are in milestone instead of divs

planatheisa:
Actually it's in complete confusion: There are div's but none of them are numbered and the subtypes are called "episode" and "choral" (it's a play). CTS is given as "book/chapter/section", none of which are ever mentioned in the text. There are milstone-tags: <milestone ed="p" n="155" unit="card"/> but the numbers coincide with the linenumbers. Also there are <sp>- and <speaker>-tags. These have to be changed, don't they? There is no link to a scan and I haven't been able to find the corresponding one. The sourceDesc is missing all references. I have not done a play in XML before and am not sure how to do this.

planatheisa:
<milestone ed="p" n="155" unit="card"/>

planatheisa:
That's what the milestone-tags look like: milestone ed="p" n="155" unit="card"

planatheisa:
@annettegessner @sonofmun

annettegessner:
This is the same problem as e.g. #103 .
Matt and I will talk about this, thanks, Tabea!

PR #218, opened 2016-06-01
#81 and #103 drama

annettegessner:
changed #81 and #103 according to #210


annettegessner:
both files passing hook test now

Issue #138, opened 2016-05-12
Only 1 citation unit in /tlg2001/tlg044/tlg2001.tlg044.opp-lat1.xml


sonofmun:

annettegessner:
there is no apparent structure in the PDF:

http://www.degruyter.com/dg/viewbookchapter.fullcontentlink:pdfeventlink/$002fbooks$002f9783110803198$002f9783110803198.267$002f9783110803198.267.pdf/9783110803198.267.pdf?format=EBOK&t:ac=books$002f9783110803198$002f9783110803198.267$002f9783110803198.267.xml

sonofmun:
The page titles looks like they show which part of the Aristotle text is basically the equivalent of what Themistius is paraphrasing on that page. My feeling is that we could perhaps put this information in milestones somewhere if we have it in the text. Otherwise the citation unit should either be lines (there are line numbers) or sentences, both of which we would need to add to the XML. I would suggest perhaps that sentences would be easier to do automatically. What are your thought @PonteIneptique ?

PonteIneptique:
- Yes for Aristotle. Maybe more in `<note type="header" corresp="">Arist.</note> or milestone.
- Why Italic ?
- Keep the linebreaks numbers in <lb n="" />
- I'd keep the paragraph as citation unit, and do create one or two new paragraphs (by showing something like `<p render="space">` [Let's find something better than that] when the original paragraph is too long.

annettegessner:
- Unfortunately we don't have the information from the page titles about the Aristotelian original in the XML. We would have to add them manually or ask somebody to do them automatically.
- Good question: I am not sure about the words in italics. Those could be translated words from the Aristotelian original, but they are very wide spread and I see no way of aligning them to the Greek original.
- Paragraph-tags should be the fasted way to segment this file and they provide quite a consistent structure. Maybe we can decide about this later, when we discussed commentaries in general a bit more.

annettegessner:
I have made every paragraph a div-chapter now. Anything else (and this) is still up for discussion.

PR #225, opened 2016-06-06
#138 first improvements

annettegessner:
I have made every paragraph a div-chapter now. Anything else (and this) is still up for discussion.


annettegessner:
passing hook test.

Please keep the issue about this file OPEN, so we can discuss it further when talking about commentaries!

Issue #93, opened 2016-05-12
Edition information for tlg2042.tlg001.perseus-grc1.xml


sonofmun:
No edition information in the header. We need to find this.

PR #226, opened 2016-06-07
#93 and #125

annettegessner:
added edition information and corrected refsDecl


annettegessner:
passing hook test now

PR #242, opened 2016-06-10
Changed XPath to delete 2 extraneous XPath, renamed file to opp-grc1.…

jduff-chs:
… Hook and Nemo pass.

### DIFF
--- a/data/tlg0645/tlg001/tlg0645.tlg001.opp-grc1.xml
+++ b/data/tlg0645/tlg001/tlg0645.tlg001.opp-grc1.xml
@@ -21,13 +21,12 @@
               <date>1911</date>
             </imprint>
           </monogr>
+          <ref target="https://archive.org/details/siustiniapologia00just">Internet Archive</ref>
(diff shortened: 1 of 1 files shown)


annettegessner:
One more thing, I forgot to mention:
Next to adding the refs target link to a scan, we would really appreciate, if you would check every files `<langUsage>` (line 35 and 36) and make sure only the main language of the text is mentioned. So in this case, please delete line 35 `<language ident="lat">Latin</language>`so that only line 36 `<language ident="grc">Greek</language>`is left. We decided to ignore things like Latin footnotes to a Greek text, so that only translations or maybe commentaries will keep both tags. Thanks!

jduff-chs:
Ah, I hadn't noticed! I'll fix that and send again, thank you for catching that. It's good to know that footnotes don't count.

annettegessner:
No problem, I myself keep forgetting to check this. ;-) Maybe these days we'll try fixing this matter for all files at once, but until then, please keep checking.

Issue #97, opened 2016-05-12
tlg2313.perseus002.perseus-grc1.xml milestones


sonofmun:
sections are in milestones instead of divs

PR #245, opened 2016-06-10
Fixed Issue #97, tlg2313.perseus002.perseus-grc1.xml

jduff-chs:
Added segs around milestones, removed two extra XPath, added work-around between segs 12 and 13 across a <p> break using <lb rend=paragraph> and <p rend=continuous>, changed XPath, edited language to grc.

This was the first commit the CHS interns worked on, at the time we did not know to make a pull request.


annettegessner:
Please delete the Latin in `langUsage` and you are good to go.
I would have chosen `<div type="textpart" subtype="section" n="1">` instead of `<seg n="1">`, but if I rememer what @PonteIneptique said correctly, this is fine too and the xPath is correct, so you can leave it like this.

PonteIneptique:
@annettegessner The point here was that we could preserve `<p>` quite easily. So in these context, if people are courageous, `<seg>` is much better.

annettegessner:
So it's better to use our div-tags, when the paragraphs end at the end of the div-tag and if we want to mark segments that end inside of a paragraph, we chose `<seg`?

annettegessner:
Ahm ... your help in merging issues is much appreciated, @sonofmun, but @jduff-chs still has to correct the langUsage and my question has not been answered yet ...

sonofmun:
Sorry, I missed the comment about the Latin language tag still in there. He can set up another PR to deal with that. And @PonteIneptique can still respond to the question even though the PR was merged.

PonteIneptique:
When possible, if there is not too much crossing between passages and p (less than 10 or 5%), seg is much better because we keep nearly all the original layout.

annettegessner:
Thanks, @PonteIneptique, I fear I might have added some closing p-tag in some files, because I didn't know how to handle this, but next time, I'll see if I can use segments.

jduff-chs:
Hello all - thank you for the clarification @PonteIneptique, and @annettegessner and @sonofmun: I will correct the langUsage tag in another PR in a few minutes, thank you for catching it!

Issue #101, opened 2016-05-12
Check citation scheme and refsDecl for /tlg0085/tlg001/tlg0085.tlg001.opp-grc3.xml


sonofmun:

planatheisa:
This is a play.

annettegessner:
As discussed: Just change the refsDecl according to #210 to just address lines and ignore all the other div-tags:
```xml
<cRefPattern n="line" matchPattern="(.+)" replacementPattern="#xpath(/tei:TEI/tei:text/tei:body/tei:div//tei:l[@n='$1'])"/>
```

jduff-chs:
Addressed in #247

PR #247, opened 2016-06-10
Issue #101 on tlg0085.tlg001.opp-grc3.xml

jduff-chs:
Corrected XPath to use lines, resolved line duplicates, corrected editor attributes, added ref.

This whole collection is a handful, because they were converted from the Perseus copy of Smyth to a scan of Sidgwick, but there are remains of Smyth numbering in spots.


annettegessner:
The ref target in line 37 should look like this:
```xml
<ref target="https://babel.hathitrust.org/cgi/pt?id=njp.32101064224940;view=1up;seq=19">HathiTrust</ref>
```
Otherwise it's fine. Thanks for taking such good care of this difficult text!

jduff-chs:
This ref syntax is now fixed, along with #248 and #249!

PR #249, opened 2016-06-10
Issue #106 on tlg0085.tlg006.opp-grc3.xml

jduff-chs:
Corrected XPath to use line, corrected editor attribute, added ref, resolved line number duplicates.


annettegessner:
The ref target in line 35 should look like this:
```xml
<ref target="https://babel.hathitrust.org/cgi/pt?id=njp.32101064224940;view=1up;seq=113">HathiTrust</ref>
```

Issue #118, opened 2016-05-12
Check citation scheme and refsDecl for /tlg1205/tlg001/tlg1205.tlg001.perseus-grc1.xml


sonofmun:

jduff-chs:
This is a fairly straightforward one, I'll start working on it, along with #120.

annettegessner:
Cool! Feel free to use the "Assignees" function on the right to assign the issue to you, just as I did right now. This makes it easier to find issues assigned to a person. Now that I have made you and your CHS colleagues members, this is finally possible.

jduff-chs:
Thank you! I've actually found that I'm no longer able to self-assign. On Friday I had noticed the option was available, but it is no longer available to any of us at CHS - I'm unsure why!

annettegessner:
Oh yeah, sorry, it seem's you can't do that due to the permission levels. My bad! I'll think about this.

PR #252, opened 2016-06-13
Issue #118, tlg1205.tlg001.perseus-grc1.xml

jduff-chs:
Removed two extra XPath, changed pb from xml:id to n.


annettegessner:
Please change in refsDecl cRefPattern n="book" to "chapter". Otherwise it looks fine.

jduff-chs:
Ok! I am never sure what to do about that n= tag. What is their purpose?

annettegessner:
The single lines of the refsDecl should represent the structure of the document, according to the div-tags, lines, paragraphs etc.. So your document has just one level, which means that all the div-tags are a child of div-edition and their subtype is "chapter". Thus the refsDecl should be
```xml
<cRefPattern n="chapter" matchPattern="(.+)" replacementPattern="#xpath(/tei:TEI/tei:text/tei:body/tei:div/tei:div[@n='$1'])"/>
```
Thanks for changing, merging now!

jduff-chs:
Thanks, that is a useful explanation!

Issue #123, opened 2016-05-12
Check citation scheme and refsDecl for /tlg1725/tlg001/tlg1725.tlg001.perseus-grc2.xml


sonofmun:

jduff-chs:
Citations and XPath are fine, runs fine in HookTest, but there are alternate page breaks that need to be converted to milestones.

jduff-chs:
This was merged, and I suppose can now be closed! @annettegessner

PR #261, opened 2016-06-14
Issue #123, marginal material in tlg1725.tlg001.perseus-grc2.xml

jduff-chs:
There were two additional page breaks: 'sec2' and 'alt'. Changed the 'sec2' pb to a seg, because it was native to the edition, and changed the 'alt' pb to milestones, because they weren't. In the process, noticed some OCR errors on koppa and sampi in a few numerals - they ended up being encoded as ?? in the text. I replaced the errors with Unicode koppas and sampis, and they work in Nemo, though they look a little clunky, see below. Is this the best practice for that problem? Also changed pb xml:id to n.
screen shot 2016-06-14 at 4 34 18 pm


annettegessner:
Looks good to me.
@PonteIneptique Please confirm, if we can leave koppa and sampi like this! Thanks!

annettegessner:
Thanks! Merged.

Issue #121, opened 2016-05-12
Check citation scheme and refsDecl for /tlg1205/tlg002/tlg1205.tlg002.perseus-grc2.xml


sonofmun:

jduff-chs:
Same structure issues as #119, milestones instead of second level reference. I'll add sections and move the <p>s inside of them.

jduff-chs:
I'm in the process of fixing this, but it's causing a strange problem in Nemo, where the sectioning becomes out of order, and two whole chapters, 4 and 5, are not recognized at all. The order becomes Chapter 1-3, Chapter 20-24, Chapter 6-19, then Chapter 25. Yet, HookTest recognizes all 25 chapters and 126 sections just fine, and passes. I've troubleshooted it a few times, even starting from scratch, but the problem in Nemo persists. Has anyone ever encountered an issue in Nemo like this?

@annettegessner @sonofmun ?

annettegessner:
I have never encountered something like this. I fear I have to direct you with Nemo questions to poor overworked @PonteIneptique again, sorry Thibault!

PonteIneptique:
Could you upload your test results ? Because these are the most important informations (first) and Nemo only comes as a checking tool :)

annettegessner:
@jduff-chs: I don't think you've made a pull request yet. Please do so, even if the XML doesn't work, so Thibault can have a look at it. Thanks!

jduff-chs:
I've made a pull request, where you can see the edits I've made, and the .zip file below contains the file in question, as well as the HookTest output, which I will also paste here:

```
Run at 2016-06-15 17:53:12
Testing /tlg1205/tlg002/tlg1205.tlg002.perseus-grc2.xml
File parsing passed
Parsed
File ingesting in MyCapytain passed
URN informations passed
urn:cts:greekLit:tlg1205.tlg002.perseus-grc2
Naming conventions passed
RefsDecl parsing passed
2 citation's level found
Passage level parsing passed
25 found
126 found
Unique nodes found by XPath passed
Available in inventory passed
Epidoc DTD validation passed
```

[issue-121.zip](https://github.com/OpenGreekAndLatin/First1KGreek/files/316804/issue-121.zip)

jduff-chs:
And here's the html from the Nemo page, you can see that things are oddly being placed out of order.


[...]

PR #262, opened 2016-06-15
Issue #121, replacing milestones with div in tlg1205.tlg002.perseus-grc2, Nemo error?

jduff-chs:
Restructured to two ref levels, put p inside sections, removed milestones, replaced pb xml:id with n. Note: Still formatting strangely in Nemo, needs more work.


sonofmun:
The file passes Hook so I will merge this PR. But the question with Nemo still remains.

Issue #134, opened 2016-05-12
Only 1 citation unit in /tlg0544/tlg001/tlg0544.tlg001.opp-grc3.xml


sonofmun:

jduff-chs:
This is quite the mystery - the only thing this file contains is the last paragraph of tlg0544.tlg001.opp-grc1, which is Mutschmann's edition of Sextus Empiricus. He notes "In fine codices (excepto M) addunt:" and adds this Ω Πυρρων speech. You can see in the picture below. opp-grc1 in our library lacks this passage (it ends with προκείμενον, the last word on the previous page), so I assume what I should do is paste this passage in at the end there?

PR #268, opened 2016-06-16
Issue #134, where to put the one paragraph from tlg0544.tlg001.opp-grc3?

jduff-chs:
The Perseus Catalog contains no mention of a third edition. Somehow, what we have as opp-grc3 contained only one paragraph, which is an closing note in Mutschmann's Sextus Empiricus, and so should be contained within to 001.opp-grc1, that edition. I've added it back to grc1. 

Below, find the image of Mutschmann's final page - currently grc1 ends after the left page, and the paragraph at the top of the right page was contained in grc3.
screen shot 2016-06-16 at 11 59 56 am


sonofmun:
Hi @jduff-chs make sure to also check the `__cts__.xml` file in the tlg0544/tlg001 folder and remove the entry for opp-grc3 since it no longer exists.

jduff-chs:
@sonofmun, thank you for catching that! Should be all good.

Issue #141, opened 2016-05-12
Only 1 citation unit in /tlg2022/tlg003/tlg2022.tlg003.opp-grc1.xml


sonofmun:

jduff-chs:
Christus Patiens. Drama, so l adjusted the citation system to use lines, there don't seem to be any other problems.

PR #271, opened 2016-06-17
Issue #141, fixing refsDecl in tlg2022.tlg003

jduff-chs:
Adjusted refsDecl to use lines (drama), seperated prologue and text body, removed lat tag, added speaker tags where needed, corrected two OCR errors, and added ref.


annettegessner:
In drama we only need one citation level, see #210. If there is no specific reason why you kept the section level, please delete it and make line the first and only citation level.

jduff-chs:
Hi Annette! I had divided it in this way because in the edition, the lines of the Prologus and the lines of the Trageodia are numbered seperately, which caused a Hook Test failure due to duplicate numbering (there is a line 1 in P and another in T, etc.). I think I could also solve this by getting rid of the Prologus, because I am unsure whether it is original to the text: it contains a Dramatis Personae and other frontmatter, it seems.

annettegessner:
I see. Works for me! If @PonteIneptique and @sonofmun don't have any objections, I'd merge this and I would like to ask you, @jduff-chs, to add this information to the "How to handle drama" issue #210.

sonofmun:
I think that if there are two sections in a document with non-sequential line numbering, then adding sections seems to be the best way to go about it. And I wouldn't delete any part of a document unless we know for sure that it is not original or, at the very least, an accepted part of the work.

PonteIneptique:
Let me get back to you, not sure we should do that :)

PonteIneptique:
I have few issues and one question : 
- I would not be against dividing but I'd prefer that we are sure this numbering specificity is common in other editions. Did you check another edition ? My quick search shows no real citation scheme as well.
- https://github.com/jduff-chs/First1KGreek/blob/3eacd3e463cc44388b61565fd149aab101f189fe/data/tlg2022/tlg003/tlg2022.tlg003.opp-grc1.xml#L86-L89 This paragraph is totally ignored by the citation scheme, when it seems to be part of the text ?
- It would be better to have `n='pr'` and `n='1'` insteand of long words.
- Another possibility is to renumber pr.1 -> pr.end as pr1 pr2 pr3 etc. Not sure Which one I like the most, it probably depends on other editions practices

jduff-chs:
Indeed, other editions number this prologue separately. See [this German blackletter from 1855](https://archive.org/stream/dietragdiechris00greggoog#page/n156/mode/2up), which seems to be one of the only other editions. That edition happens to omit both the ignored first paragraph you mentioned (which seems like an addition by the these editors given that it is unnumbered in their edition) and the Dramatis Personae (which is also unnumbered in the source text for this edition, though I numbered it in the file so that it would be included). Perhaps it is best then to leave those parts (DP and first paragraph) unnumbered so as not to make them arbitrarily part of the text, but instead make them a note or similar?

Given that the other edition also splits, either the present division or a renumbering of the form "p1" etc could work, so long as we represent them as the numbering here, which seems to be standard, without causing duplicate citations. Agreed on `n="pr"` and `n="1"`, if we stick with the section approach.

PonteIneptique:
Thanks @jduff-chs for the details.
Is the paragraph also in the other edition, if yes, we need to give it a number. We have a way to deal with different tags that @annettegessner  or @sonofmun will be able to explain to you (I have no example at hand). Dramatis Personae are never part of the scheme and should never be (we think about it as metadata, not data)

jduff-chs:
The paragraph is not in the 1855 blackletter edition (and ours is a later edition from 1885).

annettegessner:
@PonteIneptique: I don't know, if this is what you mean by handling different two tags, but if you have differently named textparts at one level, 
e.g.: in one work the 3-level-structure is EITHER book, poem, line OR book, chapter, section, the xPath should be as in this CSEL-example stoa0149b.stoa008.opp-lat1:
```xml
<refsDecl n="CTS">
        <cRefPattern n="section or line" matchPattern="(.+).(.+).(.+)" replacementPattern="#xpath(/tei:TEI/tei:text/tei:body/tei:div[@type='edition']/tei:div[@n='$1']/tei:div[@n='$2']/tei:div[@n='$3'])"/>
        <cRefPattern n="chapter or poem" matchPattern="(.+).(.+)" replacementPattern="#xpath(/tei:TEI/tei:text/tei:body/tei:div[@type='edition']/tei:div[@n='$1']/tei:div[@n='$2'])"/>
        <cRefPattern n="book" matchPattern="(.+)" replacementPattern="#xpath(/tei:TEI/tei:text/tei:body/tei:div[@type='edition']/tei:div[@n='$1'])"/>
      </refsDecl>
```

PonteIneptique:
That's the thing I meant :)

annettegessner:
Yay! Thanks for asking about this, it gave me the possibility to check and normalize how we did this in CSEL. :-)

annettegessner:
I added this to my little but growing [wiki-entry "Things to remember"](https://github.com/OpenGreekAndLatin/First1KGreek/wiki/Things-to-remember), which you all are invited to correct and expand!

jduff-chs:
That's a great trick to know with the 'or'! Should I apply this here in order to display this paragraph, or is it fair to assume it is an editorial addition/metadata? It's a synopsis, basically: "The lines of Greg. the theologian about the savior our Lord Jesus Christ and the student Judas and how he betrayed him with a treacherous kiss." or similar. Seems like it could conceivably be part of the original manuscript,  but doesn't seem to be part of the *text* proper.

PonteIneptique:
Let's open an issue specifically for that, and try to find a specialist 
/ have a scholar debate. In the end, it should not prevent this PR to be 
merged :)

jduff-chs:
More evidence (I should have thought to look at the footnote!) the editor writes "Titulum edo ex Μ, ubi rubris literis scriptus est." (I take the title from M, where it is written in red letters) in reference to that paragraph. Perhaps it should be separated as a header?

PonteIneptique:
Yes, this is indeed the case. Could you actually put the note in a note 
tag as well if it is not done ? (I have really bad internet access)

jduff-chs:
That sounds good, I'll put it in a note, and perform the other edits discussed (shorter section names)! Thanks for all the help and clarification!

jduff-chs:
Should be all set now!

annettegessner:
merging PR and closing issue 141

Issue #274, opened 2016-06-21
drop @type="edition" in refs decl


annettegessner:

sonofmun:
There needs to be a check when doing this to make sure that there is not another <div> child of <body> besides the @type="edition". Otherwise it could mess up the citations.

annettegessner:
6 files, 9 replacements
hook doesn't show any deviations in 719 success files

PR #278, opened 2016-06-28
deleted unnecessary xPath type edition

annettegessner:


annettegessner:
6 files, 9 replacements
hook doesn't show any deviations in 719 success files

annettegessner:
see issue #274

Issue #136, opened 2016-05-12
Only 1 citation unit in /tlg0732/tlg005/tlg0732.tlg005.opp-grc1.xml


sonofmun:

planatheisa:
So, doing another document from the same type, I realise that the sections, which I here numbered consecutively, could also be numbered after the "p. somenumber a/b someothernumber" structure, which comes at the beginning of every section except the very first. I am not a hundred percent sure that these don't appear more than once but I haven't found any double occurrences. Should I redo these then, do I leave it? It would probably make sense to wait if the other document passes hook (this one does).

annettegessner:
I would indeed prefer the way you numbered the other file with the "p. number-letter-number".

planatheisa:
I changed it, and now hook has found duplicates, and complains about "forbidden characters". What could these be? Not sure what to do about the duplicates. @annettegessner

planatheisa:
Could the forbidden characters be the spaces? I can remove them without problems.

planatheisa:
It's not the spaces. @sonofmun ?

sonofmun:
The forbidden character is the period "."
Since periods are used to separate citation levels, e.g., 1.1.1, they can't be used in the citation units themselves. You need to delete those, either leaving only, e.g., "p24b12", or completely removing the "p", leaving only, e.g., "24b12". If all of the citation units start with "p", I would suggest just deleting the "p" from all of them and calling them "p" or "page" in the @n attribute on the `<cRefPattern>` in the refsDecl.

PR #283, opened 2016-07-08
Tabea

planatheisa:
issue #136 added sections


annettegessner:
This file is passing the hook test, but it doesn't find the sections correctly:

RefsDecl parsing passed
1 citation's level found (CORRECT)
Passage level parsing passed
1 found (NOT CORRECT, SHOULD BE ABOUT 347)

Please check this when committing! 
I am not sure about the reason of this, please let me know, if you find out!

Also please use the other way of numbering, described in #136.

sonofmun:
The file now does not pass Hook. There can be no periods in the @n attributes. You need to remove these. For this file, I would simply delete all "p." from the @n, e.g., change "p.24a10" to "24a10". Please make sure you look at Hook and change your files when there is a problem, @planatheisa

planatheisa:
Right now, everything fails. No idea why. Still working on it.

planatheisa:
So, should anyone read this, it failed again but now i know what and will fix it.

planatheisa:
Still working on it. There are duplicates in the numbering, and I'm looking for a way to distinguish them. Apparently, brackets are also forbidden characters.

sonofmun:
Would it work to add an "a" or another letter to the duplicates to distinguish them?

planatheisa:
Did that just now.

planatheisa:
Yay. passes now.

sonofmun:
It passes now. I can't merge from my phone but I will as soon as I get to a computer.

PR #287, opened 2016-07-13
Corrected OCR in Homilia 1 of 016.

jduff-chs:
Found a handful of incorrectly OCR'd sections in 016, and figured it would be a good idea if it was as clean as possible for the purposes of the automatic alignment we'll be trying. I've corrected (by hand) all the errors in the first homily, but there exist a large number more throughout the document, which is worth noting. It seems the bilingual text was a bit of a problem, as well as significant use of &lt; and &gt; as editorial marks.


jduff-chs:
Note: In the first commit, I had find and replaced "ahi" to "alii" (a common error from this text, for some reason), but this had accidentally turned "trahit" into "traliit" and similar. The second commit corrects those accidents.

Issue #289, opened 2016-07-14
Encode critical punctuation in tlg2018.tlg010.opp-grc1


jduff-chs:
All &lt; and &gt; need to be encoded as `<add></add>`. Make sure to check xml carefully against source. See comments in #286.

sonofmun:
@jduff-chs Is this issue fully resolved now?

jduff-chs:
Yep! Closing.

PR #309, opened 2016-07-18
Issue #289, encoded critical punctuation in tlg2018.tlg010.opp-grc1

jduff-chs:
Encoded "< text >" as `<add>text</add>`, "[ text ]" as `<del>text</del>`, and "[ ]" as `<gap reason="omitted"/>`. I also corrected the OCR of the quotation punctuation »«. 

These are Greek fragments, and the editors have included references to the portions of the full Syriac text to which they correspond. We have a German translation of that full Syriac text (also from these volumes) in the corpus as tlg2018.tlg010.opp-ger1 (there is only a TLG number for the fragmentary Greek text). Should I seek to encode the references to the Syriac text in some way? URN seems to be not ideal, because the fragments and the full text share the same work number, but have different citation schemes. (Now that I think about it, the difference in citation systems alone is probably an issue, if we want all editions to be accessible by the same canonical citation scheme for the work.) @annettegessner, @sonofmun, @PonteIneptique?


jduff-chs:
For reference, the citation of opp-grc1 is just numbered fragments:

```xml
<cRefPattern n="fragment" matchPattern="(.+)" replacementPattern="#xpath(/tei:TEI/tei:text/tei:body/tei:div/tei:div[@n='$1'])"/>
```

Meanwhile, the full text in opp-ger1 is a standard book/chapter outline (though it's presently only books - the chapters weren't separated into divs by the OCR, and will also need to be converted from Roman to Arabic numbering.)

```xml
 <cRefPattern n="book" matchPattern="(.+)" replacementPattern="#xpath(/tei:TEI/tei:text/tei:body/tei:div/tei:div[@n='$1'])"/>
```

The references within the text of grc-1 to the full text are formatted as follows:
 ```xml
<head>Aus 1.23.</head>
``` 
where that refers to a fragment that corresponds to a portion of book 1, chapter 23 in the German translation (of the full Syriac text).

sonofmun:
@jduff-chs Sorry to have taken so long to get to this. This is a tough one. I think I would add a @corresp attribute and then put in what we see in the `<head>` tags as the correct attribute there. So, e.g., 

    <div type="textpart" subtype="fragment" n="3" corresp="tlg2018.tlg010:3.41-3.62">

The URN in the @corresp attribute could also include the full edition level URN of the Greek file (i.e., tlg2018.tlg010.grc-1:3.41-3.62), but I think that might be overly specific.

jduff-chs:
@sonofmun, this seems reasonable! I'll follow that example, and also fix up the German document so it has the appropriate XPath available.

Issue #95, opened 2016-05-12
tlg2255.perseus001.perseus-grc1.xml milestones


sonofmun:
the sections are in milestones instead of divs

PR #314, opened 2016-07-27
issue #95, redid it

planatheisa:
Still have to check Hook.


planatheisa:
I wanted to add a link to scan, but I couldn't find one on either InternetArchive or HathiTrust.

planatheisa:
It doesn't pass, but I can't see why. I used the file from the data folder and i checked with the other file from the same book, which passes, but I can't see any significant differences in the header beyond indentation. @sonofmun

sonofmun:
I think first that there should be two `<div>` tags in the cRefPattern, i.e., it should be 

    replacementPattern="#xpath(/tei:TEI/tei:text/tei:body/tei:div/tei:div/tei:p/tei:seg[@n='$1'])"

Try that and see if it fixes all of the problems.

planatheisa:
Thanks, that worked. It passes now.

sonofmun:
Hmm, but now it only finds 1 citation unit. Looking at the file quickly, it looks like each numbered `<seg>` elements is the only child element of a `<div>` element with the same @n attribute. Go through the file and see if this is always the case. If it is, then we can refer to the `<div>` elements instead of the `<seg>` elements. The replacementPattern would look like this:

    replacementPattern="#xpath(/tei:TEI/tei:text/tei:body/tei:div/tei:div[@n='$1'])"

If this still passes, check the more detailed results. It should find ~21 citation units.

planatheisa:
It finds the div's but I overlooked two. Correcting that now.

planatheisa:
OK. Now it finds them all.

Issue #157, opened 2016-05-12
Only 1 citation unit in /tlg4102/tlg006/tlg4102.tlg006.opp-grc1.xml


sonofmun:

annettegessner:
very similar to #158

https://archive.org/stream/catenaegraecorum02cramuoft#page/414/mode/2up

PR #316, opened 2016-07-28
issue #157

planatheisa:
First check with Hook.


planatheisa:
Duplicate references, still working on it.

sonofmun:
I think the best way is to split them and make them two separate citation levels. They refer to the page and the line number on which the text appears. So make the first number a `<div>` and call it @subtype="page" and the second number a `<div>` as the child and make it @subtype="line". I honestly don't know if that matches with the way someone might want to use this text, but it makes sense from a technical and also from a certain humanities point of view.

planatheisa:
Seems a bit of an overkill for text, that is often only a single word, but it makes sense.

planatheisa:
Not passing yet, duplicates and one forbidden character which I must have missed.

planatheisa:
It passes and finds everything, but I'm not entirely sure that it looks how it's supposed to be.

sonofmun:
Now I see what you were talking about, @planatheisa I think that what you did with the TOC is overkill. I would just put all of the `<p>` tags that contain the Greek numerals into one single `<div>`. Then it would look like:

```
<div type="commentary" subtype="TOC" n="1">
    <p>α᾿. Ὅτι οἱ μὲν ἄλλοι Εὐαγγελισταὶ ἐξ αὐτῶν ἄρχονται τῶν ὁ δὲ Ἅγιος Λουκᾶς καὶ προοιμίοις κέχρηται.</p>
    <p>β΄. Περὲ Ζαχαρίου τοῦ πατρὸς τοῦ προδρόμου.</p>
    <p>γ΄. Περὶ τοῦ εὐαγγελισμοῦ.</p>
...
```

I agree with what you said now that I understand what you meant: these do not all need to be in their own `<div>` tags. 
I like how the rest of the text looks, though.

planatheisa:
Alright

planatheisa:
Should be fine now.

sonofmun:
This looks good.

PR #317, opened 2016-07-28
Issue #290, chapter divisions and encoding critical punctuation in tlg2018.tlg010.opp-ger1

jduff-chs:
The OCR didn't originally catch the sections below the book level. I added those divs, and numbered them in arabic instead of the original roman. I also disambiguated the marginal numbering schemes, encoded the critical punctuation, and caught a lot of OCR errors, though there are plenty more in here - the ends of lines were often cut off. It now is citable so that tlg010.opp-grc1 can reference it though, which was the goal.

### DIFF
--- a/data/tlg2018/tlg010/tlg2018.tlg010.opp-ger1.xml
+++ b/data/tlg2018/tlg010/tlg2018.tlg010.opp-ger1.xml
@@ -75,6 +75,7 @@
 </profileDesc>
 <revisionDesc>
   <change who="Jack Duff" when="2016-06-21">Corrected metadata, expanded pb, removed lat and grc tag and replaced with ger, added ref, corrected book numbering.</change>
+  <change who="Jack Duff" when="2016-07-28">Completed chapter numbering, changed roman num. to arabic, OCR correction, encoding critical punctuation.</change>
 </revisionDesc>
 </teiHeader>
   <text>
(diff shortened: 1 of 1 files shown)


sonofmun:
You need to add another `<cRefPattern>` in the refsDecl to catch the chapter-level citation units. It should look like this:

    <cRefPattern n="chapter" matchPattern="(.+).(.+)" replacementPattern="#xpath(/tei:TEI/tei:text/tei:body/tei:div/tei:div[@n='$1']/tei:div[@n='$2'])"/>

Put this before the book-level cRefPattern that is already there. That should work. Give it a try.

jduff-chs:
My apologies, absolutely forgot that crucial step! I'll upload the corrected version shortly.

jduff-chs:
That commit should take care of that! I also used `<corr>` and `<supplied>` to encode more forms of critical punctuation here. `<supplied reason="unspecified">` is a bit unwieldy, but these are supplied for clarity and there is no other acceptable value for `@reason` that conveys this (the only others are "explanation", "lost", "omitted", and "subaudible").

(Also, this is our last day working here at the CHS - @josble, @caitlinmiller and I are very grateful for all the help and guidance you've all (@sonofmun, @PonteIneptique, and @annettegessner)  given us, as well as simply the opportunity to work beside you on such an important project. We've drafted up everything we've learned in [a guide](https://docs.google.com/document/d/1DCXD71l6NZUlMX3DZ7KwyvhDfYJkV-lIXnei962QT9c/edit?usp=sharing), and will be providing it to our successors, in the hope that they will learn from our mistakes! We look forward to following the progress of the project, and please let us know if we can assist with anything else in the future.)

PR #517, opened 2016-08-04
#303: added quote

planatheisa:
First upload, checking hook.


planatheisa:
Looks good. The text contains only one fragment.

PR #518, opened 2016-08-04
#304: changed <> to <add></add>

planatheisa:
First upload, checking hook.


sonofmun:
@PonteIneptique Check this one out. From the look of the `<cRefPattern>` I would think that Hook should be able to find all of the fragments. But it only finds the first 88. The problem is that there is another level between the `<div subtype="book">` element and the `<div subtype="fragment">` element for all fragments after @n="88", but I would think that the // in the `<cRefPattern>` would take care of this. Is this only Hook acting crazy or is this a real problem?

sonofmun:
Nautilus (and thus Nemo) finds all the fragments, so I am merging this.

Issue #319, opened 2016-07-29
/tlg0057/tlg067/tlg0057.tlg067.1st1K-grc1.xml


sonofmun:
>>>>> Passage level parsing failed
>>>>>> 2 found
>>>>>> 29 found
>>>>>> 0 found
>>>>> Epidoc DTD validation failed
>>>>>> error: attribute "name" not allowed here [In (L64 C57); (L166 C57); (L345 C57); (L403 C57); (L424 C57); (L482 C57); (L525 C57); (L573 C57); (L598 C57); (L740 C59); (L816 C59); (L862 C59); (L928 C59); (L942 C59); (L971 C59); (L1365 C59); (L1564 C59); (L1722 C59); (L1841 C59); (L1952 C59); (L2211 C59); (L2260 C59); (L2363 C59); (L2422 C59); (L2526 C59); (L2724 C60); (L2953 C60); (L3028 C60); (L3151 C60)]

planatheisa:
There are name-attributes in the div-elements, which are not allowed. Do I just delete them? They contain the "n" of "book" and the number of the chapter. The chapters are otherwise numbered consecutively, so you divs like: `<div type="textpart" subtype="chapter" n="25" name="II 9">`. Maybe it would be better to chenge the numbering of the chapters. Still looking for a scan to see what it looks like there.

planatheisa:
Scan can be found here: [https://catalog.hathitrust.org/Record/009723957](url). It's Volume 11.

planatheisa:
So, the chapters are not numbered consecutively in the scan (see [https://babel.hathitrust.org/cgi/pt?id=hvd.hxpp8s;view=1up;seq=83)](url)) therefore I suggest to change n-attributes for the chapters accordingy and delete the invalid name-attributes.

sonofmun:
I agree with this. Remove the @name attributes and change the chapters' @n attributes to match the scan.

planatheisa:
Do I need to change the cRefPatterns? @sonofmun

PR #523, opened 2016-08-05
#319: deleted name-attributes, corrected CTS

planatheisa:
First upload, checking hook.


planatheisa:
Passing and looking good.

Issue #326, opened 2016-07-29
/tlg0057/tlg076/tlg0057.tlg076.1st1K-grc1.xml


sonofmun:
>>>>> Passage level parsing failed
>>>>>> 10 found
>>>>>> 68 found
>>>>>> 0 found
>>>>> Epidoc DTD validation failed
>>>>>> error: attribute "name" not allowed here [In (L64 C57); (L131 C57); (L1068 C57); (L1325 C57); (L1555 C57); (L1680 C57); (L1744 C57); (L1809 C57); (L1839 C57); (L2510 C59); (L2538 C59); (L3847 C59); (L4459 C59); (L4618 C60); (L5949 C60); (L6234 C60); (L6606 C59); (L6868 C59); (L6917 C59); (L6987 C59); (L7053 C59); (L7215 C59); (L7322 C59); (L7433 C59); (L8589 C59); (L8862 C58); (L9191 C58); (L9247 C58); (L9779 C58); (L9884 C58); (L10741 C59); (L10966 C59); (L11069 C59); (L11258 C59); (L11385 C59); (L11455 C59); (L11934 C59); (L12107 C59); (L12706 C59); (L13120 C60); (L13396 C60); (L14388 C60); (L14638 C60); (L14947 C60); (L15310 C60); (L15515 C61); (L15727 C61); (L16006 C61); (L16579 C61); (L16792 C61); (L17004 C61); (L17235 C61); (L17399 C61); (L17637 C61); (L17704 C62); (L17870 C59); (L18044 C59); (L18459 C59); (L18666 C59); (L19135 C59); (L19520 C59); (L19657 C59); (L19705 C59); (L19764 C59); (L19792 C60); (L19823 C58); (L20043 C58); (L20516 C58)]
>>>>>> error: text not allowed here [In (L5936 C3); (L13948 C3); (L18701 C3); (L20444 C3)]

planatheisa:
There is text from two different volumes in there: Book 1-6 are from volume 12, book 7-10 from volume thirteen. The header contains only volume 12, books 1-6. Can I just add the other volume and books?

sonofmun:
If you mean add it to the header, then yes. Just add volume 13 and the pages that it covers there.

planatheisa:
What do I do about the pagenumbers? They give no information about the volume.

sonofmun:
[...]

PR #528, opened 2016-08-08
#326: corrected divs and CTS, changed pagenumbers and expanded header

planatheisa:
First upload, checking hook.


planatheisa:
Looking good.

Issue #340, opened 2016-07-29
/tlg0057/tlg040/tlg0057.tlg040.1st1K-grc1.xml


sonofmun:
>>>>> Passage level parsing failed
>>>>>> 0 found
>>>>>> 0 found
>>>>>> 0 found

PR #535, opened 2016-08-08
#340: corrected CTS

planatheisa:
First upload, checking hook.


planatheisa:
Looks good. Contains only one book and no additonal structure.

Issue #469, opened 2016-07-29
/tlg0627/tlg035/tlg0627.tlg035.1st1K-grc1.xml


sonofmun:
>>>>> Passage level parsing failed
>>>>>> 4 found
>>>>>> 0 found
>>>>>> 0 found

PR #538, opened 2016-08-08
corrected cRefPattern and langUsage

annettegessner:


annettegessner:
passing hook test, 
close issue #469, when merging

Issue #468, opened 2016-07-29
/tlg0057/tlg009/tlg0057.tlg009.1st1K-grc1.xml


sonofmun:
>>>>> Passage level parsing failed
>>>>>> 3 found
>>>>>> 21 found
>>>>>> 0 found
>>>>> Epidoc DTD validation failed
>>>>>> error: attribute "name" not allowed here [In (L64 C57); (L96 C57); (L267 C57); (L368 C57); (L586 C57); (L669 C57); (L936 C57); (L997 C57); (L1098 C57); (L1343 C59); (L1446 C59); (L1874 C59); (L1994 C59); (L2144 C59); (L2368 C59); (L2856 C60); (L3036 C60); (L3169 C60); (L3302 C60); (L3626 C60); (L3805 C60)]

PR #539, opened 2016-08-08
#468 deleted name-attributes, corrected cRefPattern and langUsage

annettegessner:


annettegessner:
passing hook, close #468

PR #540, opened 2016-08-08
#467

annettegessner:
corrected cRefPattern and langUsage, deleted name-attributes, corrected section numbering


annettegessner:
passing hook

Issue #353, opened 2016-07-29
/tlg0057/tlg001/tlg0057.tlg001.1st1K-grc1.xml


sonofmun:
>>>>> Passage level parsing failed
>>>>>> 14 found
>>>>>> 0 found
>>>>>> 0 found
>>>>> Epidoc DTD validation failed
>>>>>> error: text not allowed here [In (L386 C3)]

PR #544, opened 2016-08-10
#353: corrected CTS

planatheisa:
First upload, checking hook.


sonofmun:
The refsDecl needs to be changed. There is only 1 citation level ("chapter")

planatheisa:
Should be fine now.

Issue #496, opened 2016-07-29
/tlg0057/tlg079/tlg0057.tlg079.1st1K-grc1.xml


sonofmun:
>>>>> Passage level parsing failed
>>>>>> 18 found
>>>>>> 0 found
>>>>>> 0 found
>>>>> Epidoc DTD validation failed
>>>>>> error: attribute "name" not allowed here [In (L556 C57)]

PR #613, opened 2016-08-15
Tabea #496

planatheisa:
First upload, checking hook.


planatheisa:
There is a chapter in this work, that says (in Greek) "chapter 6 and 7". How can I put this in the `@n` of the div? Putting `n="6-7"` doesn't work (should have known that...). See: [https://babel.hathitrust.org/cgi/pt?id=hvd.hxpp8u;view=1up;seq=245](url) for reference. @annettegessner @sonofmun

sonofmun:
This is a really good question. Any ideas @PonteIneptique ? I will bring it up at the OGL meeting this afternoon.

PonteIneptique:
I can't see the original unfortunately. Make it 6 for the moment if you need. Best thing would be to search where the division takes place in other editions :)

sonofmun:
The weird thing is that what follows is a single sentence in Greek in square brackets. It's almost like it isn't part of the text as it exists but a summary of these two sections from somewhere else. I don't know enough about Galen to know how to deal with this.

annettegessner:
Matt and I just opened a new issue to discuss things like that: see #614 Use of <div type="commentary" ...>

planatheisa:
I changed it to "6" and it passes now.

PR #661, opened 2016-08-21
#502: corrected divs and CTS, added reference and expanded header

planatheisa:
First upload, checking hook.


planatheisa:
Looks good.

Issue #342, opened 2016-07-29
/tlg0627/tlg050/tlg0627.tlg050.1st1K-grc1.xml


sonofmun:
>>>>> Passage level parsing failed
>>>>>> 18 found
>>>>>> 0 found
>>>>>> 0 found

PR #666, opened 2016-08-21
#342: corrected CTS, added reference

planatheisa:
First upload, checking hook.


planatheisa:
First upload, checking hook.

PR #715, opened 2016-09-06
Added the Lake version of Eusebius Ecclesiatical History. 

sonofmun:
The English needs to be separated from the Greek and put in its own file (tlg2018.tlg002.opp-eng1).


sonofmun:
The file passes Hook. I am going to merge my own PR so that I can add an issue to separate the English and the Greek.

PR #658, opened 2016-08-19
Issue #399

LucieSty:
added ref url, changed lang tags, removed two levels of XPath, removed initial paragraph numbering, added vol.8 to all pb elements

PR #717, opened 2016-09-06
fix xpath

LucieSty:
Update ref expression to chapter to match text


sonofmun:
@LucieSty So this change took. I can see it in the new file. Now, on this same branch, make the change to the `<lang>` that @annettegessner pointed out (https://github.com/OpenGreekAndLatin/First1KGreek/pull/658) then push and it should be added to this pull request. Let me know if you have problems.

sonofmun:
Looks good. Merging now.

PR #718, opened 2016-09-07
Change resp

sonofmun:
I only changes the `<respStmt>`s for the texts that originated in Leipzig and from the Graeco-Arabic project. More conversion will be need for many of the older files from Perseus to make them convertible as a batch.


sonofmun:
It didn't break anything so it can be merged. This may require that Lucy's outstanding PR might have to be merged with this one by hand (which I can do). We'll see.

sonofmun:
This needs to be merged so that no conflicts occur with other files so I will do it myself.

Issue #633, opened 2016-08-17
Fix tlg0646.tlg004.1st1K-grc1.xml


sonofmun:
Testing /tlg0646/tlg004/tlg0646.tlg004.1st1K-grc1.xml
File parsing failed
File ingesting in MyCapytain failed
URN informations failed
Naming conventions failed
RefsDecl parsing failed
Passage level parsing failed
Unique nodes found by XPath failed
Available in inventory failed
Epidoc DTD validation failed
fatal: XML document structures must start and end within the same entity. [In (L61 C44)]

PR #719, opened 2016-09-07
First try to fix tlg0646.tlg004.1st1K-grc1.xml. 

sonofmun:
The conversion from Beta Code had choked in the middle. The full text should now be present.


sonofmun:
File now passes. Can be merged. Close #633 when merging.

PR #720, opened 2016-09-28
Tlg5026.tlg007

sonofmun:


sonofmun:
Now passing Hook. Take a look and merge if it looks OK, @annettegessner

sonofmun:
@PonteIneptique Didn't you say that Marianne was waiting for the Scholia in the Odyssey? If so, merge this PR and they are there!

PonteIneptique:
I certainly hhave a look

sonofmun:
Enclosed the first section of the epigraph in its own <div>. Still passing Hook. I have also sent instructions to DDD to correct the notes. I think that it would be OK to merge it in its present form. What do you think, @PonteIneptique ?

PR #727, opened 2016-10-05
#364: corrected CTS and divs, added reference

planatheisa:
First upload, checking hook.


planatheisa:
Looks good.

planatheisa:
In reference to the scan, there seems no need to have more than one level for the CTS.

PR #733, opened 2016-10-05
Added '.DS_Store' (Mac useless unimportant metadata) to .gitignore

planatheisa:
These .DS_Store files showed up today. Never seen them before, internet says they are metadata for/from Finder like window-size, position and other really unnecessary information. Suggestions were to add them to .gitignore. Hope that helps.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 .directory 
+.DS_Store


sonofmun:
Great! Thanks for this, @planatheisa .

annettegessner:
:-)

PR #775, opened 2016-10-11
Update tlg0018.tlg015.opp-grc1.xml

AlisonBabeu:
Fixed last issue with header.


sonofmun:
@AlisonBabeu There was a merge conflict on this pull request. I fixed it, but I am not entirely sure what happened. It may have been that you made changes to this file earlier, in a different pull request, and then made more changes, but using the old file, not the one that you already made changes on. I don't know if this is the case, but if it is, it was probably caused by you making a branch from master for this pull request before the master branch had been updated with your previous changes. Does that sound possible? I am just stabbing in the dark, here.

AlisonBabeu:
I'm pretty sure there that what you described is exactly what happened, totally, definitely! :)

Issue #339, opened 2016-07-29
/tlg0645/tlg002/tlg0645.tlg002.perseus-grc2.xml


sonofmun:
>>>>> Naming conventions failed
>>>>> RefsDecl parsing passed
>>>>> Passage level parsing failed
>>>>>> 15 found
>>>>>> 0 found
>>>>>> 0 found

planatheisa:
@sonofmun  Need the correct header for this file.

sonofmun:
The `<respStmt>` should look like this. Is there anything else that looks wrong in the headers?

```
<funder>A Google Digital Humanities Award</funder>
<principal>Gregory Crane</principal>
<respStmt>
    <persName xml:id="DDD">Digital Divide Data</persName>
    <resp>Corrected and encoded the text</resp>
</respStmt>
<respStmt>
    <persName>Gregory Crane</persName>
    <resp>Editor-in-Chief, Perseus Digital Library</resp>
</respStmt>
<respStmt>
    <persName>Matt Munson</persName>
    <resp>Project Manager (University of Leipzig), 2016 - present</resp>
</respStmt>
<respStmt>
    <persName>Annette Gessner</persName>
    <resp>Project Assistant (University of Leipzig) 2015 - present</resp>
</respStmt>
<respStmt>
    <persName>Thibault Clérice</persName>
    <resp>Lead Developer (University of Leipzig)</resp>
</respStmt>
<respStmt>
    <persName>Bruce Robertson</persName>
    <resp>Technical Advisor</resp>
</respStmt>
```

planatheisa:
publication statement looks like this:
`<publicationStmt>
                <publisher>Trustees of Tufts University</publisher>
                <pubPlace>Medford, MA</pubPlace>
                <authority>Perseus Project</authority>
 </publicationStmt>`

Not sure if that is correct. @sonofmun

sonofmun:
Yes, keep this as it is. This text was first prepared for Perseus at Tufts so we should leave it like this.

planatheisa:
Ok.

PR #789, opened 2016-10-14
#339: corrected CTS and langUsage, added reference, updated __cts__.xml

planatheisa:
First upload, checking hook.


planatheisa:
Looks good.

Issue #425, opened 2016-07-29
/tlg4027/tlg001/tlg4027.tlg001.opp-grc1.xml


sonofmun:
>>>>> Passage level parsing failed
>>>>>> 3 found
>>>>>> Duplicate references found : 3

PR #798, opened 2016-10-16
#425: corrected CTS, divs and langUsage

planatheisa:
First upload, checking hook.


planatheisa:
Duplicates...

planatheisa:
Looks good.

sonofmun:
There is something strange with the citation scheme here. Hook says that it finds 5 books and 9 chapters in this work. But in the XML I only find 1 book. Can you take a look and see what is going one here, @planatheisa ? Check the XML tree to see what it might be finding instead of the `<div type="textpart" subtype="book">` as the first level of citations.

planatheisa:
It finds the commentary-divs. Is it not supposed to?

sonofmun:
No, it should not find the ones in commentary since that is not a part of the original text (I assume, anyway). Try changing both replacement patterns so that they have div[@type="edition"], i.e., 

    replacementPattern="#xpath(/tei:TEI/tei:text/tei:body/tei:div[@type='edition']/tei:div[@n='$1'])"

and

    replacementPattern="#xpath(/tei:TEI/tei:text/tei:body/tei:div[@type='edition']/tei:div[@n='$1']/tei:div[@n='$2'])"

planatheisa:
Duplicates...

planatheisa:
Do I number the commentary-divs?

sonofmun:
Hmm, the `<div type="commentary">` elements are inside the `<div type="edition">`. I don't think this should be the case. @PonteIneptique what is your opinion here?

PonteIneptique:
`<div type="commentary">` sibling of `<div type="edition">`. Then it could have four divs `<div type="commentary">` which are numbered

sonofmun:
Right, so move the `<div type="commentary">` elements up so that they are siblings (instead of children) of `<div type="edition">`. Then, with @type="edition" in the replacementPattern, it should hopefully work.

planatheisa:
Seems to be fine now.

planatheisa:
Seems to be fine now.

PR #803, opened 2016-10-19
Added tlg0641.tlg001.First1K-grc1 Babrii Fabulae Aesopeae

sonofmun:


sonofmun:
Passes Hook. @annettegessner please check and merge.

annettegessner:
@sonofmun Here the langUsage has to be corrected and there is only one level of citation with one occurrence, which looks not too good. Will there be an issue for this file, one it's uploaded?

sonofmun:
Looking at the XML, it appears that there actually are citation units that were marked as `<lb>` by DDD at the paragraph level. I will need to look at it more closely next week when I get back from Prague. This might be a text we need to send back.
See #812 for my comment on langUsage.

PR #804, opened 2016-10-19
Added tlg1765.tlg003.First1K-grc1.xml Vita Aesopi Pl vel Accursiana

sonofmun:


sonofmun:
Passes Hook. Please check and merge @annettegessner

sonofmun:
Still passes. @annettegessner Notice on this one that, since the citation scheme often overlapped paragraphs, that what I did to deal with this was to make the `<p>` elements all empty (i.e., `<p/>`) and then enclosed all of the text in `<ab>` elements. That way the paragraphs are still marked but there are no overlapping tags. We will need to think about whether this is the best solution or if we should do something else. If you have an opinion on this @PonteIneptique , I would be interested to hear it.
A good example of the problem is https://github.com/OpenGreekAndLatin/First1KGreek/blob/8e99f234fb0ad774b78045c13963932229ff6958/data/tlg1765/tlg003/tlg1765.tlg003.1st1K-grc1.xml#L2070-L2077

PonteIneptique:
Can you tell me why are you using AB O_O ?

sonofmun:
Because of this definition in the TEI guidelines:

    <ab> (anonymous block) contains any arbitrary component-level unit of text, acting as an anonymous container for phrase or inter level elements analogous to, but without the semantic baggage of, a paragraph.

It needs to be something that contains the text but is not marked at all in the edition. Is there a better choice than `<ab>` for this?

PonteIneptique:
Ok fair enough, weird but fair enough.
And then, why the do you have p ? If you want to show paragraph, do not use empty P but complete P. You can as well use `@render="continuous"` or something.

sonofmun:
So would this mean closing the `<p>` tag before the new `<div>`, and then opening a new `<p render="continuous">` in the new `<div>`?

PonteIneptique:
I guess :) Or make lb that have a special render parameter ;)

sonofmun:
Neither of us are TEI gurus (you more than me, however), so I think I will use @rend on the `<p>`. Check out here: http://www.tei-c.org/release/doc/tei-p5-doc/de/html/ref-att.global.rendition.html

    @rend indicates how the element in question was rendered or presented in the source text.

Does this make sense? 
I am still not sure I understand why this is better than an empty `<p>` with `<ab>` (or something else) inside of it. But I don't really understand the semantics of TEI all that well and perhaps `<p>` as a milestone (i.e., empty tag) makes absolutely no sense.

PonteIneptique:
`perhaps <p> as a milestone (i.e., empty tag) makes absolutely no sense.`
-> This is it :D Sorry, I should have said that from the beginning. Grumpy morning :/

sonofmun:
Good. Then I will use `<p rend="continuous">`.

PonteIneptique:
Sorry man, it actually should be p rend="merge" as per the practice of PerseusDL. I misremembered.

sonofmun:
OK. I will change it.

sonofmun:
Still passing. Can now be merged.

PR #805, opened 2016-10-19
Added tlg1765.tlg004.First1K-grc1.xml Vita Aesopi (fort. auctore Apht…

sonofmun:
…honio)

### DIFF
--- a/data/tlg1765/tlg004/__cts__.xml
+++ b/data/tlg1765/tlg004/__cts__.xml
@@ -1,0 +1,7 @@
+<ti:work xmlns:ti="http://chs.harvard.edu/xmlns/cts" groupUrn="urn:cts:greekLit:tlg1765" xml:lang="grc" urn="urn:cts:greekLit:tlg1765.tlg004">
+  <ti:title xml:lang="lat">Vita Aesopi (fort. auctore Aphthonio)</ti:title>
+  <ti:edition urn="urn:cts:greekLit:tlg1765.tlg004.First1K-grc1" workUrn="urn:cts:greekLit:tlg1765.tlg004">
+    <ti:label xml:lang="lat">Vita Aesopi (fort. auctore Aphthonio)</ti:label>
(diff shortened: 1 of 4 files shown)


sonofmun:
Passes Hook. Please check and merge, @annettegessner

sonofmun:
tlg1765.tlg004.1st1K-grc1.xml still passes. Can be merged.

annettegessner:
Just one div, but that should be ok. LangUsage should only be Greek.

sonofmun:
Since @xml:lat is used to define certain elements in the metadat(`<name xml:lang="lat">ALFREDI EBERHARD</name>`), I think it should be defined in `<langUsage>`. Or is this not the case? The language of the edition itself is defined in the `<div type="edition">` tag. What is your view on this @PonteIneptique ? If @xml:lang="lat" is not defined in langUsage, where should it be defined?

annettegessner:
Very good question! I thought LangUsage should be used to describe, which language is used in the text body, not necessarily elements in the header, but now we have the information in the edition-div, I am not sure. This might have to be changed for a lot of files...

sonofmun:
Reading the information on @langUsage here http://www.tei-c.org/release/doc/tei-p5-doc/en/html/HD.html#HD41 suggests that we don't have to define accepted language tags, like "lat" or "grc". See this sentence:

A language element may be supplied for each different language used in a document. If used, its ident attribute should specify an appropriate language identifier, as further discussed in section vi.1. Language Identification. This is particularly important if **extended language identifiers** have been used as the value of xml:lang attributes elsewhere in the document.

I am not sure what best practice is here but I think at the very least, we should have a `<language>` element for every language we have defined in the document or the header with an @xml:lang attribute. This wouldn't be too much of a problem to do automatically. But we should decide if we want to do this before we go through the trouble. This discussion has been moved to Issue #821 .

sonofmun:
For now, I would suggest merging if `<langUsage>` is the only unresolved problem. Then I can deal with that en masse depending on our decision there.

annettegessner:
Ok, I'll merge all PRs with only LangUsage being unclear!

PR #806, opened 2016-10-19
Added tlg1765.tlg005.First1K-grc1.xml Βιβλίον μυθικὸν τοῦ Αἰσώπου

sonofmun:

PR #807, opened 2016-10-19
Added tlg3118.tlg001.First1K-grc1.xml Ιστορικον Συντιπα του Φιλοσοφου

sonofmun:
I added a citation level that was not in the volume at the level of the <p> tags, otherwise there would have been only a single citation level for a 136 page work.


sonofmun:
Passes now. @annettegessner Take a look and, if you agree with adding the citation level at the paragraph, feel free to merge. Otherwise we can have a conversation about this.

sonofmun:
tlg3118.tlg001.1st1K-grc1.xml still passes. Can be merged.

annettegessner:
tlg3118.tlg001.1st1K-grc1.xml looks good, only Lang Usage should be Greek only.
There is also tlg1765.tlg005.First1K-grc1.xml in this PR.

sonofmun:
The second file was from another pull request that somehow snuck into this PR. It shouldn't cause a problem and should not be evaluated here (PR #806 adds this text). See Issue #821 for langUsage.

PR #808, opened 2016-10-19
Added tlg3118.tlg002.First1K-grc1.xml Narrationes de Syntipa

sonofmun:
I added a citation level that wasn't in the edition at the level of the `<p>` tags. Otherwise this would have been a 60 page work with no citation scheme.


sonofmun:
Passes now. @annettegessner Take a look and, if you agree with adding the citation level at the paragraph, feel free to merge. Otherwise we can have a conversation about this.

sonofmun:
tlg3118.tlg002.1st1K-grc1.xml still passes. Can be merged.

PR #809, opened 2016-10-20
Tlg5022.tlg002

sonofmun:
Corrected URN (from First1K to 1st1K)


sonofmun:
tlg5022.tlg002.1st1K-grc1.xml still passes. Can be merged.

annettegessner:
tlg5022.tlg002.1st1K-grc1.xml looks good, LangUsage should be Greel only.

PR #810, opened 2016-10-20
Tlg4075.tlg002

sonofmun:
Corrected URN (from First1K to 1st1K)


sonofmun:
tlg4075.tlg002.1st1K-grc1.xml still passes. This can be merged.

annettegessner:
LangUsage (depending on Thibault's answer) Greek only(?) and we should check for another possibility to structure. Maybe paragraphs?

sonofmun:
I think the text is short enough that we probably can go with only one citation unit. I would prefer not to impose our own citation scheme unless it is absolutely necessary. langUsage #821

PR #811, opened 2016-10-20
Added tlg1220.tlg001.1st1K-grc1.xml Batrachomyomachia

sonofmun:


sonofmun:
Passes. Can be checked and merged @annettegessner

annettegessner:
Check LangUsage and check for other possibility to structure, but careful, there are paragraphs and lines in this work.

sonofmun:
I have no idea what to do with the citation scheme on this document. It is the "Batrachomyomachia". It is in Homeri Opera so perhaps @Eumaeus could help us here. Chris, do you have any suggestions here?

lmuellner:
I asked a friend of mine who has just prepared a new edition of the Batrach. if there is any other citation scheme for the work than its line numbers — which I seriously doubt, for what it’s worth.

lmuellner:
His answer is that the line numbers in Allen (Homeri Opera) are standard, though there are Byzantine “interpolations” that mess up the numbers. Is that helpful?

sonofmun:
I will have to look more closely at the text next week. I seem to remember that there are lines that are interrupted by paragraphs. The problem is how to deal with this situation.

sonofmun:
I think I have fixed this one, @planatheisa HookTests is only showing 8 lines when there should be 303, but I think this is a problem with the way HookTexts finds the citation units. Since there is an unnumbered `<p>` element between the previous `<div>` and the `<l>` elements, I think Hook only finds the lines under the first `<p>` and then craps out. Can you confirmed this @PonteIneptique ?

PonteIneptique:
Normally this bug has been resolved. Have you tried the GetValidReff ?

PonteIneptique:
Few notes : 
- there is a line which is not covered by the XPath (l[n="117"]) because it is not in a p.
- What is the reason behind p's ?
- There is numerous line with space in them :/

sonofmun:
The paragraphs are in the text itself, but the accepted citation scheme are the lines in this edition. I thought about deleting the `<p>` elements, but figured they should stay since they are in the text itself.
Ah, the @n attributes sometimes have spaces. Yes, I brought that directly over and will need to change it.
And I will put line 117 in a p element.

PonteIneptique:
I have all 303 Lines otherwise. Wonder if something went wrong with the version of Hook...

sonofmun:
I have noticed this before that, I think, when there is an element with no @n attribute between two levels of the citation hierarchy that Hook just looks for the elements under the first of those mid-level elements, like here it looks only under the first `<p>`.

PonteIneptique:
The thing is it is a resolved bug. I am watching the situation right now in the HookWorker and it seems the version of Capitain was somehow downgraded... I am trying to fix this.

sonofmun:
I restarted the Hook worker (hookworkerrq) last Weds. or Thurs., one of the days you were off last week. But I wouldn't think that would downgrade the CapiTainS version.

PonteIneptique:
I am cleaning up the whole environments

sonofmun:
This should be ready to check and merge now @planatheisa

planatheisa:
I just merged a PR on tlg1765.tlg005.First1K-grc1.xml. Why is this here also? Also there is xml:lang="eng"/"lat" in the sourceDesc of tlg1220.tlg001.1st1K-grc1.xml (and "lat" in langUsage).

sonofmun:
Hmm, I think I will remove that file from this PR so that it doesn't overwrite any changes that we made in the other PR. So hold on just a bit on this one.

sonofmun:
OK, so I merged tlg1765.tlg005 from the master branch, so the version here should be up-to-date with the version that we merged already. You should probably check it just to make sure. And I corrected the language tags on tlg1220

PR #812, opened 2016-10-20
Added tlg1252.tlg002.1st1K-grc1.xml Certamen Homeri et Hesiodi

sonofmun:


sonofmun:
Passes Hook. Please check and merge @annettegessner

annettegessner:
Is this file Latin and Greek? If not, please change langUsage. I see only Greek.

sonofmun:
@xml:lang="lat" is used in the header (e.g., `<listBibl xml:lang="lat">`) and so needs to be defined in `<langUsage>` as well.

PR #813, opened 2016-10-20
Add metadata

LucieSty:
we changed the title statement and source desc in the header. we
modified the language id. we were unable to provide a reference to the
original target.


planatheisa:
@sonofmun This is a file I worked on, and which you changed also. Do I just close it?

sonofmun:
Yes, I don't think there is anything else we can do with it.

Issue #631, opened 2016-08-17
Fix stoa0121.stoa001.opp-grc1.xml


sonofmun:
Testing /stoa0121/stoa001/stoa0121.stoa001.opp-grc1.xml
File parsing passed
Parsed
File ingesting in MyCapytain failed
Text can't be read through Capitains standards
URN informations failed
Naming conventions failed
RefsDecl parsing failed
Passage level parsing failed
Unique nodes found by XPath failed
Available in inventory failed
Epidoc DTD validation failed
error: element "TEI.2" not allowed anywhere [In (L2 C8)]

planatheisa:
@sonofmun Would this be the same header as #339 ?

sonofmun:
Yes

PR #816, opened 2016-10-24
#631: corrected CTS, <gap>, divs and langUsage, added reference, and …

planatheisa:
...`__cts__` files

First upload, checking hook.


planatheisa:
Not quite right yet.

planatheisa:
one more time

Issue #632, opened 2016-08-17
Fix tlg0610.perseus002.perseus-grc1.xml


sonofmun:
Testing /tlg0610/perseus002/tlg0610.perseus002.perseus-grc1.xml
File parsing passed
Parsed
File ingesting in MyCapytain failed
Text can't be read through Capitains standards
URN informations failed
Naming conventions failed
RefsDecl parsing failed
Passage level parsing failed
Unique nodes found by XPath failed
Available in inventory failed
Epidoc DTD validation failed
error: value of attribute "type" is invalid [In (L43 C27); (L54 C27); (L65 C27); (L75 C27); (L84 C27); (L92 C27); (L106 C27); (L119 C27); (L130 C27); (L144 C28); (L157 C28); (L172 C28); (L179 C28); (L191 C28); (L204 C28); (L216 C28); (L230 C28); (L242 C28); (L255 C28); (L265 C28); (L276 C28); (L285 C28); (L298 C28); (L308 C28); (L321 C28); (L335 C28); (L344 C28); (L351 C28); (L370 C28); (L380 C28); (L390 C28); (L406 C28); (L419 C28); (L431 C28); (L441 C28)]

planatheisa:
@sonofmun Would this be the same header as #339 ?

sonofmun:
Yes

PR #817, opened 2016-10-24
#632: corrected CTS, divs and langUsage, added reference

planatheisa:
First upload, checking hook.


planatheisa:
"Available in inventory failed"
Where is the error here? @sonofmun

sonofmun:
That means that the URN in the `__cts__.xml` file is not the same as the one in the XML file itself. You should also make sure that both of these match up with the filename for the XML file.

planatheisa:
Looks good.

Issue #634, opened 2016-08-17
Fix tlg1271.tlg001.1st1K-grc1.xml


sonofmun:
Testing /tlg1271/tlg001/tlg1271.tlg001.1st1K-grc1.xml
File parsing passed
Parsed
File ingesting in MyCapytain failed
Text can't be read through Capitains standards
URN informations failed
Naming conventions failed
RefsDecl parsing failed
Passage level parsing failed
Unique nodes found by XPath failed
Available in inventory failed
Epidoc DTD validation passed

planatheisa:
Would this be the same header as #339 ? @sonofmun

sonofmun:
Yes.

planatheisa:
Also, there are bible passages given in the margins, and in the English translation there are quotation marks, however it's not so easy to determine where exactly the quote starts in the Greek text. I could compare with the bible text, but that would take a while. There are over a hundred passages cited. Also, they might not be cited exactly, so more of an indirect quote. Do i put a `<q type="mentioned">` there for the section? What if there are two "quotes" in a section? See: [https://archive.org/stream/apostolicfather00lakegoog#page/n24/mode/2up](url) for reference. @sonofmun  @annettegessner

sonofmun:
A quick look at the XML tells me that these were marked up as `<bibl>`. Since these are, in essence, "bibilographic citations", that works as it is. The TEI guidelines also say that "the sub-components may or may not be explicitly tagged." That means that we can just leave it like it is and it means, "Around in this area of the text is a reference to this biblical verse." So I think that we can leave them as they are here. Any more effort might actually misinterpret what the editor meant with the note. And, as you say, it would be too much work, at least at this point.

planatheisa:
Do I add a "corresp" to "bibl"? That would not take as long.

planatheisa:
Although, I'm not sure how to do multiple citations. @sonofmun

sonofmun:
I don't think adding a @corresp attribute would add much value. And it might actually introduce problems, for instance with multiple citations. And we have the citation in the text of the `<bibl>` element already. What are your thoughts about the advantage of @corresp, @planatheisa ?

planatheisa:
Same as with any quotation, you can look at the original text. Determine more correctly what is meant and in what context maybe to interpret the text. @sonofmun

planatheisa:
[...]

PR #818, opened 2016-10-24
#634: corrected header, CTS, divs and langUsage, added reference and …

planatheisa:
…corresp

First upload, checking hook.


planatheisa:
Looks good. I also added a cts-file for both the tlg1271 and tlg1271.tlg001.

PR #819, opened 2016-10-24
#635: corrected header, CTS, divs and langUsage, added reference, cor…

planatheisa:
…resp and the cts file

First upload, checking hook


planatheisa:
Looks good.

Issue #636, opened 2016-08-17
Fix tlg1311.tlg001.1st1K-grc1.xml


sonofmun:
Testing /tlg1311/tlg001/tlg1311.tlg001.1st1K-grc1.xml
File parsing passed
Parsed
File ingesting in MyCapytain failed
Text can't be read through Capitains standards
URN informations failed
Naming conventions failed
RefsDecl parsing failed
Passage level parsing failed
Unique nodes found by XPath failed
Available in inventory failed
Epidoc DTD validation passed

planatheisa:
@sonofmun  @annettegessner  This does not give an author. Worldcat says 

> 4. THE 'DIDACHE' (Teaching) is a book of moral precepts and of precepts for church life.

Do I put Anonymus there?

sonofmun:
Yes.

PR #820, opened 2016-10-24
#636: corrected CTS, divs and langUsage, added reference, corresp and…

planatheisa:
… cts files

First upload, checking hook.


planatheisa:
Looks good.

PR #822, opened 2016-11-01
Added tlg0708.tlg001.1st1K-grc1.xml

sonofmun:
Ammonius Qvi Dicitvr Liber De Adfinivm Vocabvlorvm Differentia


sonofmun:
@PonteIneptique @annettegessner Apparently @cause="omitted" on the `<add>` element is an invalid value in EpiDoc. I have changed them all to @cause="fix", since that is apparently what happened here, i.e., because the text was omitted the editor added it in order to fix the problem. The question is whether we want DDD to use this value from now on or @cause="unclear", which seems to be the other valid value.

sonofmun:
Passes Hook now. If the solution with the `<add cause="fix">` is acceptable (and there are no other errors), feel free to merge @annettegessner

annettegessner:
Looks good to me and I am fine with your fix so I'll merge.

PR #824, opened 2016-11-01
tlg0363.tlg007-Need to fix character error for title in   __cts__.xml

AlisonBabeu:
So I have proposed a change here, but I'm not sure this is technically correct.  I was wondering what I should do about the HTML character entity here for the equals sign, which wasn't correct anyway since it should have been &#61.  I have copied in in the proper TLG title here: Apotelesmatica (= Tetrabiblos) since I'm not sure how to represent the = sign in the title.


lcerrato:
Hi, @AlisonBabeu 
Could you edit the title to reflect the related author/work name? Such as adding (tlg0363.tlg007) to the beginning of this.

lcerrato:
FWIW, I had a work file unhappy with `&` (ampersand) whereas regular UTF-8 chars should work just fine. 
It failed the hook tests, so that was my indication that it was unhappy with the ampersand.

sonofmun:
It passes now so I am going to merge it.

PR #828, opened 2016-11-02
Added tlg0082.tlg001.1st1K-grc1.xml

sonofmun:
Apollonius, Dyscolus, De constructione


sonofmun:
There is still a small problem here but I am going to merge this so that the UVA interns can work on it.

PR #831, opened 2016-11-03
Added tlg0082.tlg001.1st1K-grc1.xml Apollonius Dyscolus De pronominibus

sonofmun:
Also moved tlg0082.tlg004.1st1K-grc1.xml to the proper folder.
Note also that I ignored the Bekker pages as citation scheme for this work and, instead, numbered the paragraphs. If anyone has major problems with this, I can certainly change it. Otherwise I would prefer to keep it like this


sonofmun:
tlg0082.tlg001.1st1K-grc1.xml now passes. I will fix tlg0082.tlg002.1st1K-grc1.xml and add it to this PR (since the incorrect version has already been added).

sonofmun:
Both files now pass Hook. Please check and merge @annettegessner

annettegessner:
I am perfectly fine with the paragraph numbering, since it's easier to read and dominant in the text anyway. Merging.

PR #838, opened 2016-11-04
Update tlg0059.tlg037.First1K-grc1.xml

AlisonBabeu:
Edited a few things in the header to standardize it with the other headers, and fixed a typo.


sonofmun:
Still passes Hook. Merging.

PR #841, opened 2016-11-07
Added tlg1210.tlg001.1st1K-grc1 Autolycus, De Sphaera Quae Movetur

sonofmun:


sonofmun:
Passes Hook. Can be checked and merged @annettegessner

PR #849, opened 2016-11-08
Added tlg1126.tlg003.1st1K-grc1.xml Anubion Fragmenta, tlg1337.tlg003…

sonofmun:
….1st1K-grc1.xml Dorotheus Fragmenta alia antiqua, and tlg2583.tlg001.1st1K-grc1.xml Manetho Apotelesmatica


sonofmun:
All files pass now. Added the "paragraph" citation level for tlg2583.tlg001.1st1K-grc1.xml in order to break the text into manageable units. Otherwise there was no obvious citation scheme.

annettegessner:
I am very uncertain about the paragraphs. It seems that the intended way of structuring this work is line numbering. This might be doable, but there are some lines out of continuous numbering, e.g. lines 180, 179 and 181 here:
https://archive.org/stream/autolycidesphaer00autouoft#page/n337/mode/2up
What do you think @sonofmun ?

sonofmun:
Lines out of sequence is no problem. I will try and take a look at this to see whether numbering the lines could be done automatically.

planatheisa:
@sonofmun  Is there a reason for tlg1126.tlg003.1st1K-grc1.xml to contain langUsage="lat"?

sonofmun:
It depends on whether we use the @xml:lang="lat" anywhere in the document. If we do, then we should define what it means in the header. If not, then there is no need to have it, although it is also not wrong if we do have it.

planatheisa:
tlg1126.tlg003.1st1K-grc1.xml : idno does not correspondend with filename, no Latin text in the textpart (so there should not be Latin in langUsage)

tlg1337.tlg003.1st1K-grc1.xml : idno, no Latin text in texpart

tlg2583.tlg001.1st1K-grc1.xml : idno, no Latin text in textpart, no divs type="paragraph" (as given in CTS) instead numbered `<p>` elements.

sonofmun:
Both tlg1337 and tlg2583 have @xml:lang="lat" in the header, so normally we would need to define the "lat" language attribute. But in both of these, the name of the editor is miscategorized as Latin, so I will change both of those.
And tlg2583 is undergoing major changes, so the last level of citation will actually be `<l>` instead of `<p>`. This file should be finished today.

sonofmun:
I think everything finally matches up with all of these files. Please check for errors and merge @planatheisa

planatheisa:
I'll merge this presently. Just another quick question. One of the document contains this: `<name xml:lang="ger">Hermann Koechly</name>`. I guess it's not really important, but should this be reflected in langUsage? As it seems it should when it's Latin?

sonofmun:
The language usage rules are really kind of weird. If it is an accepted language code, like "ger", then you don't need to use it. "lat" is also an accepted code and so we really don't even need that. I think "grc" is less well recognized and so should be included, however. Of course, now I can't find the documentation that declares this. So I guess I will go ahead and add it, just to be safe.

sonofmun:
I just deleted the @xml:lang attribute on the editor tag. We never declare German or any other modern languages and, if it isn't necessary, I would prefer not to start.

PR #851, opened 2016-11-08
Added tlg0358.tlg001.1st1K-grc1.xml Nichomachus of Gerasa Introductio…

sonofmun:
… arithemica and tlg0358.tlg005.1st1K-grc1 UNKOWN Problemata arithemica
This second work is simply tacked on to the end of the edition of Nichomachus. It apparently comes from one of the manuscripts of the Introductio by Nichomachus, as the title to the edition states, "Accedunt Codicis Cizensis Problemata Arithemetica". I don't know who this work if from, so I have simply put it under Nichomachus for now. Suggestions as to what to do with this @AlisonBabeu ?


sonofmun:
Both texts pass.

AlisonBabeu:
I'm not entirely sure. I did a search in the TLG for this text and didn't find it, which of course doesn't mean its not in there, but that does mean we don't have a ready author/anonymous ID to use for a CTS URN. I haven't been in the habit of creating ID for anonymous Greek works in general.

sonofmun:
Started an issue to fix this problem (#887). I think for now we should merge this and deal with that text later.

planatheisa:
@sonofmun I am never sure: How important is it for the `<idno type="filename">` to contain the actual filename?

sonofmun:
I am not sure what it is used for but it is in our schema, so we should make sure that we have it in every document. If you notice that there is a PR where there is no `<idno>`, then have the author change it.

planatheisa:
Will you change it or should I?

sonofmun:
I will change it for this PR.

sonofmun:
Still passes Hook. Please check and merge @planatheisa

Issue #829, opened 2016-11-02
Fix tlg0082.tlg004.1st1K-grc1.xml


sonofmun:
Passage level parsing failed
4 found
582 found
Duplicate references found : 2.68

decline93:
Went into this file to check the structure. Everything seemed fine (four books, seemingly 582 sections, and no duplicate reference at 2.68), so I added lang tags in the appropriate places and ran it through Travis, which it passed completely. Should we close out this issue and should I create a pull request to update the lang tags, etc?

sonofmun:
Yes. Thanks for finding this, @decline93

PR #856, opened 2016-11-15
Fixed header for tlg0082.tlg004

AlisonBabeu:
Updated header information for this work.


sonofmun:
This work still needs to be fixed Issue #829 . But I will merge this PR with these fixes.

Issue #648, opened 2016-08-17
Fix tlg5034.tlg001.perseus-grc1.xml


sonofmun:
Testing /tlg5034/tlg001/tlg5034.tlg001.perseus-grc1.xml
File parsing passed
Parsed
File ingesting in MyCapytain failed
Text can't be read through Capitains standards
URN informations failed
Naming conventions failed
RefsDecl parsing failed
Passage level parsing failed
Unique nodes found by XPath failed
Available in inventory failed
Epidoc DTD validation failed
fatal: file not found: http://www.tei-c.org/Guidelines/DTD/tei2.dtd [In ]

Issue #647, opened 2016-08-17
Fix tlg4170.tlg001.perseus-grc1.xml


sonofmun:
Testing /tlg4170/tlg001/tlg4170.tlg001.perseus-grc1.xml
File parsing passed
Parsed
File ingesting in MyCapytain failed
Text can't be read through Capitains standards
URN informations failed
Naming conventions failed
RefsDecl parsing failed
Passage level parsing failed
Unique nodes found by XPath failed
Available in inventory failed
Epidoc DTD validation failed
fatal: file not found: http://www.tei-c.org/Guidelines/DTD/tei2.dtd [In ]

planatheisa:
@sonofmun 
This has no publicationStmt. What do I add there?

sonofmun:
@AlisonBabeu @lcerrato What XML markup do the entities in this file represent? Or where can they be found? The entities are:

```
&responsibility;
&fund.AnnCPB;
&Perseus.publish;
```

The file is here: https://github.com/OpenGreekAndLatin/First1KGreek/blob/master/data/tlg4170/tlg001/tlg4170.tlg001.perseus-grc1.xml

planatheisa:
The same goes for the issue #648, which, incidentally comes from the same book.

lcerrato:
@sonofmun @planatheisa 
We use the entities transformation scenario to expand these:
https://github.com/PerseusDL/tei-conversion-tools/tree/master/xslt

They are also visible here 
https://github.com/PerseusDL/tei-conversion-tools/blob/aa48fedc3fab954f13bfdca16d97e38c1d5f6747/dtd/PersTeiHead.dtd

I ran a transform:

&responsibility; 

     <sponsor>Perseus Project, Tufts University</sponsor>
          <principal>Gregory Crane</principal>
		<respStmt>
		<resp>Prepared under the supervision of</resp>
		<name>Lisa Cerrato</name>
		<name>William Merrill</name>
		<name>Elli Mylonas</name>
		<name>David Smith</name>
		</respStmt>`
	
&fund.AnnCPB; 
        
         <funder n="org:AnnCPB">The Annenberg CPB/Project</funder>
   
&Perseus.publish;

	<publicationStmt>
		<publisher>Trustees of Tufts University</publisher>
		<pubPlace>Medford, MA</pubPlace>
		<authority>Perseus Project</authority>
		</publicationStmt>

I expect this is perhaps a cut and paste of a header and may not be applicable to this text.

[...]

PR #865, opened 2016-12-01
#648: corrected CTS, LangUsage, and divs, added publicationStmt

planatheisa:
First upload, checking hook.


planatheisa:
duplicates, working on it

planatheisa:
Still not passing.

planatheisa:
And still not passing.

planatheisa:
No idea where the problem lies here. I thought it was the urn in the >__cts__.xml< file, but it must be something else. Also, I couldn't find a scan. It looks as if it would be the same as #647, but the text doesn't match.

planatheisa:
Well, it was the urn I guess. Passing now.

PR #866, opened 2016-12-01
Update header for stoa0146d.stoa001.opp-grc1.xml

AlisonBabeu:
I made a few changes to the header to make them standard with other headers, such as not
having everything in capital letters.


sonofmun:
This file has other problems but I will merge these changes.

PR #870, opened 2016-12-05
Standardized header for tlg0085.tlg002.opp-grc3.xml

AlisonBabeu:
Deleted "machine readable text" as subtitle for header consistency and added in editor information.


sonofmun:
Adding the `<editor>` tag in line 31 broke Epidoc compliance here. It needs to be outside the `<imprint>` element.

AlisonBabeu:
Yeah I noticed that and am pretty sure I put it in the right place for all the other files. Do I need to fix this and submit a new pull request?

Issue #882, opened 2016-12-06
Update metadata for tlg5034.tlg001.perseus-grc4.xml


sonofmun:
The `<ti:description>` in the work level `__cts__.xml` file should be: Scholia in Pythian Odes
The `<ti:title>` for all editions here should be "Scholia in Pindarum (scholia vetera)"

sonofmun:
Change this: this should have a different work-level URN and, thus, will need to have the proper changes made. It should be tlg5034.tlg001b.perseus-grc1, the `<ti:title>` should be "Scholia in Pindarum Pythian Odes" and the description should be "Scholia in Pindarum Pythian Odes, Drachmann, Teubner, 1910"

PR #889, opened 2016-12-12
#651, #882: corrected header and divs, added CTS, reference, langUsag…

planatheisa:
…e, cts file

First upload, checking hook.


planatheisa:
Looks good. @sonofmun What happens to tlg5034.tlg001.opp-grc4? That should be obsolete now.

sonofmun:
Yes, go ahead and delete the whole tlg5034/tlg001 folder, so that we only have the four folders that end with the letters 1-d there.

PR #907, opened 2016-12-13
Header corrections for tlg027.tlg048.opp-grc2.xml

AlisonBabeu:
Made edits to the headers with some various traumas on the way.


planatheisa:
@sonofmun Shouldn't it be `<biblScope unit="volume">1</biblScope>` instead of `<biblScope unit="volume">vol. i.</biblScope>`?

sonofmun:
I think so. @AlisonBabeu could you make that change as well?

AlisonBabeu:
I've made the change and updated the file. Should be all set to go. I also realized the volume should be 2 not 1 but the format should be correct.

PR #913, opened 2016-12-20
Edited header and cts_xml file for tlg0553.tlg001.1st1K-grc1.xml

AlisonBabeu:
Fixed typo in author name in cts_xml file and updated the header.


sonofmun:
@AlisonBabeu We usually try to only change one file per pull request. That makes it easier to figure out precisely what is going wrong and what to fix. It also allows us to get the files that are compatible merged into master more quickly.

Issue #346, opened 2016-07-29
/tlg0057/tlg059/tlg0057.tlg059.1st1K-grc1.xml


sonofmun:
>>>>> Passage level parsing failed
>>>>>> 4 found
>>>>>> 64 found
>>>>>> 0 found
>>>>> Epidoc DTD validation failed
>>>>>> error: attribute "name" not allowed here [In (L64 C57); (L157 C57); (L217 C57); (L264 C57); (L559 C57); (L596 C57); (L654 C57); (L707 C57); (L772 C57); (L888 C59); (L929 C59); (L968 C59); (L1000 C59); (L1016 C59); (L1064 C59); (L1399 C59); (L1524 C59); (L1565 C59); (L1597 C59); (L1623 C59); (L1640 C59); (L1659 C59); (L1675 C59); (L1691 C59); (L1704 C59); (L1777 C59); (L1805 C59); (L1847 C59); (L1866 C59); (L1898 C59); (L2069 C59); (L2085 C59); (L2129 C59); (L2287 C59); (L2438 C59); (L2591 C59); (L2851 C59); (L3124 C59); (L3291 C59); (L3377 C60); (L3606 C60); (L3804 C60); (L3899 C60); (L4073 C60); (L4306 C60); (L4354 C60); (L4688 C60); (L4881 C59); (L4968 C59); (L5422 C59); (L5545 C59); (L5576 C59); (L5691 C59); (L5729 C59); (L5780 C59); (L5890 C59); (L5942 C60); (L6096 C60); (L6184 C60); (L6228 C60); (L6236 C60); (L6260 C60); (L6285 C60); (L6302 C60)]
>>>>>> error: element "table" not allowed anywhere [In (L311 C8); (L314 C35); (L333 C8); (L355 C8); (L380 C8); (L401 C8); (L422 C8); (L443 C8); (L465 C8); (L490 C8); (L1105 C8); (L1122 C8); (L1139 C8); (L1164 C8); (L1184 C8); (L1204 C8); (L1224 C8); (L1247 C8); (L1267 C8); (L1287 C8); (L1307 C8); (L1310 C35); (L1328 C8); (L3152 C8); (L3169 C8); (L3186 C8)]

sonofmun:
Tables could be a problem. See #251

Pria20:
There are no tables in this file of the Medicorum graecorum. Was that part of the document moved into a separate file?

sonofmun:
[...]

Issue #251, opened 2016-06-13
How to handle figures, tables et al.?


annettegessner:
In case of a work including some kind of figure or table, at this point we have to do the following:
Mark the textpart as a figure or [table](http://www.tei-c.org/release/doc/tei-p5-doc/fr/html/ref-table.html). Make sure you preserve the text inside the figure or table as accurately as possible and if necessary add a link to the exact PDF-page, where it can be found.

Example for figure with uncorrected text:
tlg4015.tlg001.opp-grc1
diagram on [page 28](https://archive.org/stream/commentariaina13pt12akaduoft#page/n49/mode/2up), 

Markup:

``` xml
<figure>
 <graphic url="https://archive.org/stream/commentariaina13pt12akaduoft#page/n49/mode/2up"/>
 <figDesc>οὐσία
συμβεβηκός
οὐκ ἐν
δποκειμένω
καὶ
καθ’ δπο
κειμενου
κα
θόλου
ἀσύστατον
μερ
ἐν ὕπο.
κειμένῳ
καὶ οὐ
καθ’ δπο-
κειμενου
ἷκόν</figDesc>
</figure>
```

BUT correct the text inside the figure (`<figDesc>`) by correcting the spelling and preserving the layout! 

At a later point, there will be [SVG](http://tei.it.ox.ac.uk/Talks/2007-10-31-MMWorkshop/P4toP5-1.xhtml) added to the XML.

PonteIneptique:
Note : SVG can be really easily created with some tools online such as https://www.gliffy.com even if it is not their primarly goal (I think libre office Draw accepts also SVG as an output)

sonofmun:
I don't think this will work for tables since, as far as I can tell (http://www.stoa.org/epidoc/gl/latest/app-elements.html) the `<table>` element is not allowed anywhere in EpiDoc. Perhaps I am missing something but this means we need to figure out a different way to deal with tables.

lmuellner:
Hey Matt, maybe this has to change. There are going to have to be tables for some authors, like Ptolemy and others. Perhaps we talk about this at the meeting tomorrow? Best, Lenny

sonofmun:
@simonastoyanova How would you handle this? E.g., isn't that big stone with all the tributes paid to Athens considered to be a table? If so, how is it encoded?

LucieSty:
Hi ,

[...]

PR #914, opened 2016-12-21
#346: did a lot of stuff ages ago, also first test of tables (#251)

planatheisa:
Checking hook.


planatheisa:
Looks good.

sonofmun:
The @ident on the language tag should be "grc" instead of "greek". Otherwise it looks good.

planatheisa:
Something happened to hook...

PonteIneptique:
Test has run a second, there is duplicate node but it should be sufficient for the moment.

sonofmun:
It looks like the file passes with the rerun test. I am merging and then I can look at the results of the new test.

Issue #361, opened 2016-07-29
/tlg0057/tlg063/tlg0057.tlg063.1st1K-grc1.xml


sonofmun:
>>>>> Passage level parsing failed
>>>>>> 28 found
>>>>>> 0 found
>>>>>> 0 found
>>>>> Epidoc DTD validation failed
>>>>>> error: element "table" not allowed anywhere [In (L184 C8); (L211 C8); (L231 C8); (L251 C8); (L274 C8); (L300 C8); (L320 C8); (L340 C8); (L360 C8)]

PR #915, opened 2016-12-23
#361: corrected lots of stuff ages ago, also: tables

planatheisa:
Checking hook.


sonofmun:
File passes on the second run so I will merge.

Issue #367, opened 2016-07-29
/tlg0057/tlg060/tlg0057.tlg060.1st1K-grc1.xml


sonofmun:
>>>>> Passage level parsing failed
>>>>>> 4 found
>>>>>> 20 found
>>>>>> 0 found
>>>>> Epidoc DTD validation failed
>>>>>> error: attribute "name" not allowed here [In (L64 C57); (L279 C57); (L504 C57); (L588 C57); (L641 C57); (L777 C57); (L866 C57); (L945 C57); (L1187 C57); (L1210 C59); (L1293 C59); (L1680 C59); (L2083 C59); (L2381 C59); (L2467 C60); (L2693 C60); (L2976 C60); (L3328 C59); (L3547 C59); (L3983 C59)]
>>>>>> error: element "table" not allowed anywhere [In (L3253 C8); (L3278 C8)]

PR #916, opened 2016-12-23
#367: corrected lots of stuff ages ago, also: tables

planatheisa:
Checking hook.


sonofmun:
File passes with the rerun test.

Issue #419, opened 2016-07-29
/tlg0057/tlg049/tlg0057.tlg049.1st1K-grc1.xml


sonofmun:
>>>>> Passage level parsing failed
>>>>>> 7 found
>>>>>> 0 found
>>>>>> 0 found
>>>>> Epidoc DTD validation failed
>>>>>> error: element "table" not allowed anywhere [In (L389 C8); (L417 C36); (L433 C14); (L508 C14); (L583 C14)]

PR #917, opened 2016-12-23
#419: corrected lots of stuff ages ago, also: tables

planatheisa:
Checking hook.


sonofmun:
File passes when rerun.

PR #918, opened 2017-01-09
Added tlg1443.tlg004.1st1K-grc1.xml Ignatius Ad Ephesios

sonofmun:


sonofmun:
Passing now (check the second result for the file on Hook).
FYI, Hook is now messed up because of a problem with the Github API. You will need to run it manually to get good results.

sonofmun:
@AlisonBabeu I created a new URN for this work because all seven letters from Ignatius have the same URN in TLG (tlg001). I am planning, instead, to create URNs for each of these letters, starting with tlg1443.tlg004 (Ad Ephesios), this file here.

planatheisa:
@sonofmun  How do I run Hook manually? I didn't know you could do that. Also: there are citations in this file but the @corresp does not contain the URN.

sonofmun:
@planatheisa You can check out how to install and run HookTest locally here: https://github.com/Capitains/HookTest
As for the @corresp attributes, did you ever set something up to automatically insert those for biblical citations? I am going to do that this morning and we can then use that to make these changes automatically.

sonofmun:
Still passes Hook (now working online)

planatheisa:
@sonofmun No, there were always so many different spellings and OCR errors that I didn't think it would save a lot of time, if any.

sonofmun:
@planatheisa I wrote something quick for biblical citations and the URNs. The point would be to correct the citations in the document first and then to run the Python script over them. I did all of these files with that script. It at least saves the time of always having to look up and type in the URN.

planatheisa:
Sounds good.

planatheisa:
@sonofmun What version of python do you need to install Hook? I have 3.4 and I get an error when I try to install it.

sonofmun:
@planatheisa does it give you an error about urllib3 and SSL content? If so, try upgrading your Python to 3.5 and see if that helps. Of course, if you have any Python projects that rely on the version of Python that you have installed, you should make sure that Python 3.4 is still available.

PR #920, opened 2017-01-09
Added tlg1443.tlg006.1st1K-grc1.xml Ignatius Ad Trallianos

sonofmun:
Checked the file locally and it passes Hook. Please check and merge, @planatheisa


planatheisa:
@sonofmun Wrong header. And again, the @corresp are missing the URN.

sonofmun:
Still passing

PR #927, opened 2017-01-10
Added tlg1487.tlg001.1st1K-grc1.xml Maximus Astrologus Maximi et Ammonis

sonofmun:


sonofmun:

planatheisa:
@sonofmun There is a lot of Latin text in the footnotes. I am not sure if that should be reflected in langUsage or not.

sonofmun:
@planatheisa We don't have an @xml:lang attribute on the notes and we never use the "lat" value for @xml:lang, so I don't think we need it here. Plus, the notes are mixed Greek and Latin and I am pretty sure that you can't have two @xml:lang attributes on a single node. So if we really wanted to mark them, we would have to mark the notes as "lat" and then every Greek word or letter in there as "grc". I don't think that is worth it.

planatheisa:
Sorry for my rather long absence. Doesn't pass Hook here because of duplicates. The chapters upwards of nr 9 have to be n+1. Also, it says that the branch has conflicts with `data/tlg1487/__cts__.xml` but I can't see what could be the reason.

sonofmun:
The reason for the conflict was, I think, extra white space in the version of the file on the master branch. I will fix the duplicates.

sonofmun:
I believe it passes now. Please check and merge, @planatheisa

PR #933, opened 2017-01-11
Added tlg0565.tlg002.1st1K-grc1.xml Soranus De signis fracturarum

sonofmun:


sonofmun:

planatheisa:
@sonofmun The idno @filename does not contain the filename. Also, CTS has "paragraph" but there is no corresponding div.

sonofmun:
@planatheisa Thanks for catching the missing filename. The "paragraph" level are in numbered `<p>` elements.

planatheisa:
OK. I always thought the names had to match up. But aparently not.

sonofmun:
As I understand it, the @n element on the cRefPattern is simply what is shown in any viewer. In this case, since the `<p>` elements represent paragraphs, the @n attribute is "paragraph". We do the same thing with lines, i.e., @n="line" matches the `<l>` elements.

PR #934, opened 2017-01-11
Added tlg0565.tlg003.1st1K-grc1.xml Soranus De fasciis

sonofmun:


sonofmun:

planatheisa:
@sonofmun There is an `xml:lang="la"` in `<listBibl>`that I don't think is necessary.

sonofmun:
@planatheisa Deleted it.

PR #935, opened 2017-01-11
Added tlg0565.tlg004.1st1K-grc1.xml Soranus Vita Hippocratis

sonofmun:


sonofmun:

planatheisa:
@sonofmun Is it `xml:lang="la"` or `xml:lang="lat"`? Because the `__cts__.xml` files all have the latter. This might be true for others as well. I haven't checked all.

sonofmun:
@planatheisa It should be "lat". I think we always try to use 3 letter codes, at least for the ancient languages.

PR #938, opened 2017-01-11
Added tlg0661.tlg002.1st1K-grc1.xml Polyaenus Macedo Excerpta Polyaeni

sonofmun:


sonofmun:

planatheisa:
@sonofmun There are two types of `<lb>` in there which are not distinguishable.

sonofmun:
I think that one represent paragraph numbers and the other represents line numbers on the page in the edition. It would be nice to be able to transform the ones representing the paragraph numbers to citation units, but they are not put in there consistently enough to do that automatically. There are some sections that have absolutely no `<lb>` representing paragraph numbers, some where all `<p>` elements have an `<lb>`, and some where some have them and some don't. I think the sections are short enough that we don't have to worry about this problem right now.

planatheisa:
idno/filename does not end in `.xml`. Seems like nitpicking but I guess it's better to be consistent with these things.

sonofmun:
@planatheisa No, I don't think that is nitpicking. I figure that if the idno is ever used, it will be used by a machine and if the .xml is missing, there will be an error. I will fix it.

sonofmun:
Fixed. Please check and merge @planatheisa

PR #953, opened 2017-01-27
Updateding work files and _cts_.xml files for some tlg0732 works by Alexander Aphrodisias

AlisonBabeu:
Also added work  title and author info to tlg0732.tlg004.opp-grc1.xml


AlisonBabeu:
I think I have fixed the one issue. Thanks for finding that @sonofmun

PR #959, opened 2017-02-14
Added Lucian of Samosata Cataplus

sonofmun:


sonofmun:

planatheisa:
`<title>` in `<biblStruct>` seems to be the authors name instead of the title.

sonofmun:
Nice catch, @planatheisa But in this case, this is actually correct. The volumes that contain these works are actually entitled "Lucian".

PR #969, opened 2017-02-14
Added Lucian of Samosata Adversus indoctum et libros multos ementem

sonofmun:


sonofmun:

PR #977, opened 2017-02-21
Added Aristotle De anima

planatheisa:
Checking Hook.


planatheisa:
I corrected the URN, but Hook has a problem with `"omitted"` in `<add cause="omitted">`. What should I do about that?

annettegessner:
In this document we have `<add cause="omitted">`and `<gap reason="omitted"/>`. Is there a difference between the two?

sonofmun:
I think `<add>` marks something that was added by the editor while `<gap>` is something that was marked as missing by the editor without offering an insertion.

sonofmun:
@reason is required by Epidoc on `<gap>`. The possible values are: ellipsis, illegible, lost, omitted.
@cause is not necessary on `<add>`, but we are adding it to give some idea of what editorial symbols are in the edition itself. The possible values for @cause are: fix, unclear. So instead of @cause="omitted", we should just generally say @cause="fix".

planatheisa:
Finally pasisng Hook now.

sonofmun:
I think we need another citation level here. The next level (probably "section") is marked by the marginal notes in this format `<note type="marginal">402a</note>`. You should be able to do this with a regex in Oxygen or any text editor. I do this with regex pretty often, so let me know if you want some advice on how to do it.

planatheisa:
The problem is, that there is no clear indicator where the section would start. There are places where the marginal not is practically in the middle of a large passage which I would have to cut in half somewhere. But on page five there is a `<lb n="2"/>` ff. which seems to be marking sections/chapters. See [http://digital.slub-dresden.de/werkansicht/dlf/109861/13/](url). I think that would make a better candidate.

PR #989, opened 2017-02-23
Added Galen Quos quibus (ap. Oribasius)

planatheisa:
Checking Hook.


planatheisa:
See pull request #977.

planatheisa:
Passing hook.

sonofmun:
I think we need another level of citation here. The next level would be marked as `<lb n="x">`. Again, I would do this with a regex. Let me know if you have questions.

planatheisa:
Looks good.

PR #993, opened 2017-02-24
Update tlg2959.tlg002__cts__.xml

AlisonBabeu:
Edited language data and added in further edition data.


planatheisa:
To be consistent I think you should change `xml:ang="eng"` to `"lat"` in the second edition as well.

sonofmun:
I agree with @planatheisa here that we should have both either as @xml:lang="eng" or @xml:lang="lat" or with no @xml:lang attribute.

PR #996, opened 2017-02-24
Update tlg2959.tlg003.opp-ger1__cts__.xml

AlisonBabeu:
Changed title language and added in edition metadata.


planatheisa:

AlisonBabeu:
I can't believe I missed updating the language on the second edition, you are quite right, the various edition titles should all have their language changed for consistency.

PR #997, opened 2017-02-24
Update header and other info tlg2959.tlg003.opp-ger1.xml

AlisonBabeu:
Updated header and bilbiographic info. Also added in German as language tag since this work is in German.

### DIFF
--- a/data/tlg2959/tlg003/tlg2959.tlg003.opp-ger1.xml
+++ b/data/tlg2959/tlg003/tlg2959.tlg003.opp-ger1.xml
@@ -61,13 +61,13 @@ License</p>
 <pubPlace>Germany</pubPlace>
 </publicationStmt>
 <sourceDesc>
-<listBibl xml:lang="ger">
+<listBibl xml:lang="de">
 <biblStruct>
 <monogr>
 <title>Methodius</title>
 <editor>
 <persName>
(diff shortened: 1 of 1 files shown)


sonofmun:
A couple more things to change. First, since @xml:lang="grc" and @xml:lang="lat" do not existing in this document, we should delete those lines in `<langUsage>` as well. Also, the @xml:lang="lat" attribute needs to be deleted from the `<editor>/<persName>/<name>` element. And, finally, since the work is in german, could you add an @xml:lang="de" attribute to the `<div type="edition">` element?

sonofmun:
Also, I just noticed that you define the @xml:lang attribute for German as "de", but then you use "ger" in the elements. I think you should change @xml:lang="ger" to @xml:lang="de".

AlisonBabeu:
Yeah I noticed that halfway through, I think it is that in the MODS records I create I have gotten used to using the language tags they use for Marc records which is still ger for German rather than de for Deutsch. I'll go through the files and change all of this for consistency.

sonofmun:
@AlisonBabeu According to @PonteIneptique, we should use the three letter language codes from ISO 693-3. That means instead of 'de', we should use 'deu'. And following up on the metadata conversation from yesterday, could you add an `@xml:lang="lat"` attribute to `<titleStmt>/<title>`?

PR #998, opened 2017-02-24
Update tlg2959.tlg004__cts__.xml

AlisonBabeu:
Updated language and added edition info.


planatheisa:

PR #999, opened 2017-02-24
Update tlg2959.tlg004.opp-ger1.xml

AlisonBabeu:
Updated various information including header, bibliographic information, etc.


sonofmun:
Go ahead and remove "Greek" and "Latin" from `<langUsage>` and add @xml:lang="de" to the `<div type="edition">` element.
Also, I just noticed that you define the @xml:lang attribute for German as "de", but then you use "ger" in the elements. I think you should change @xml:lang="ger" to @xml:lang="de".

sonofmun:
As in #997 "de" should be "deu". And could you add `@xml:lang="lat"` to `<titleStmt>/<title>`?

PR #1001, opened 2017-02-24
Update tlg2959.tlg005__cts__.xml

AlisonBabeu:
Edited language and added in edition information.

### DIFF
--- a/data/tlg2959/tlg005/__cts__.xml
+++ b/data/tlg2959/tlg005/__cts__.xml
@@ -1,7 +1,7 @@
 <ti:work xmlns:ti="http://chs.harvard.edu/xmlns/cts" groupUrn="urn:cts:greekLit:tlg2959" xml:lang="grc" urn="urn:cts:greekLit:tlg2959.tlg005">
-  <ti:title xml:lang="eng">De Creatis (Fragmenta Ap. Photium, Bibl. Cod. 235)</ti:title>
+  <ti:title xml:lang="lat">De Creatis (Fragmenta Ap. Photium, Bibl. Cod. 235)</ti:title>
   <ti:edition urn="urn:cts:greekLit:tlg2959.tlg005.opp-grc1" workUrn="urn:cts:greekLit:tlg2959.tlg005">
-    <ti:label xml:lang="eng">De Creatis (Fragmenta Ap. Photium, Bibl. Cod. 235)</ti:label>
-    <ti:description xml:lang="eng">Methodius, De Creatis (Fragmenta Ap. Photium, Bibl. Cod. 235)</ti:description>
+    <ti:label xml:lang="lat">De Creatis (Fragmenta Ap. Photium, Bibl. Cod. 235)</ti:label>
+    <ti:description xml:lang="eng">Methodius, De Creatis (Fragmenta Ap. Photium, Bibl. Cod. 235), Bonwetsch, Hinrichs, 1917</ti:description>
   </ti:edition>
 </ti:work>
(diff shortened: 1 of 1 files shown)


planatheisa:

AlisonBabeu:
I am less certain on some of these files in terms of changing all of the  `xml:lang="eng"` to "lat" only in terms of the final section of the edition metadata.  For example, this line: ` <ti:description xml:lang="eng">Methodius, De Creatis (Fragmenta Ap. Photium, Bibl. Cod. 235), Bonwetsch, Hinrichs, 1917</ti:description>` contains both Latin author and title name, but the editors and publisher names are not Latin, of course they are also not English either. Any thoughts, @sonofmun  on this? And thanks so much @planatheisa for catching this inconsistency on my part so we can figure out how to proceed.

sonofmun:
I expect that we could simply remove the @xml:lang attributes and be OK here for precisely the reason you mention, @AlisonBabeu . If I recall correctly, if the language is obvious and common, then the @xml:lang attribute is not necessary. I think that is the case here, though I am open to be contradicted by someone who knows more than I do.

AlisonBabeu:
@sonofmun I'm not entirely certain how and when it is decided to add the `xml:lang` attribute and to be perfectly honest this was the first time that I changed that attribute on some kind of consistency whim, since most of the cts_xml files I'm pretty sure have "eng" as the default language for titles that are in Latin. My guess would be the language attribute would become important if we started having lots of editions and were using the cts_xml files to keep track of all of the title variants.  Should I just ignore that language attribute from now on in all the cts_xml files, or perhaps should we just ignore it on the edition metadata line since that line is typically multiple languages?

sonofmun:
Since my plan is to automatically create the `__cts__.xml` files from the metadata in the teiHeader of the files, I think we should just leave it off. Then, if we want correct language tags on the title, etc., then we need to have them correctly on the elements in the teiHeader. So if a title is in Latin, then we should have an @xml:lang="lat" attribute on the `<title>` element. Then I can just automatically extract that and put it in the information for that edition. If there is no @xml:lang attribute, then I will not add one to the `__cts__.xml` file.

sonofmun:
@AlisonBabeu @planatheisa @PonteIneptique So, after doing some research into language tags, it looks like we need to make some changes to how we encode things. First, a default language tag should be given for a node that will describe all of the language of it and its descendents. We do this on the `<div type="edition">` element, but I think we need to do it for the `<teiHeader>` as well, setting the value to @xml:lang="en". Then, when anything is different from that, we need to explicitly declare it. So if we have a `<title>` or `<author>` in Latin, we need to add the @xml:lang="la" attribute to those elements. Third, notice the "la" for Latin. That is what http://www.iana.org/assignments/language-subtag-registry/language-subtag-registry says is the accepted tag for Latin. If we would prefer to use "lat", then we should define it under `<langUsage>/<language>`. And, finally, since 'grc' is the accepted language tag for "Ancient Greek (to 1453)", I don't think we need to define it in `<langUsage>/<language>`.
Perhaps all of these conversations have already been had but, just in case they haven't, this seems to me to conform to the standards of how languages are encoded. Comments and critique are welcome here.

PonteIneptique:
Hey : while I partially agree with the use of a general xml:lang="eng" we decided to take the road of the norm 639-3 which has the quality of always being the 3 chars long. Please keep with this system, it's always possible to convert from this one to another if need be.

PonteIneptique:
And remember that ti:edition, ti:translation, ti:work and ti:textgroup node should not bear an xml:lang about the language expressing their metadata but rather the language of the text (for edition, translation and work) and not have one for textgroup.

sonofmun:
That makes sense. I should be able to automatically extract that information from the @xml:lang attribute on the `<div type="edition">` element.

PR #1002, opened 2017-02-24
Update tlg2959.tlg006__cts__.xml

AlisonBabeu:
Edited language data and added edition info.


planatheisa:

sonofmun:
See discussion on #1001

PR #1005, opened 2017-02-24
Update tlg2959.tlg007.opp-grc1.xml

AlisonBabeu:
Updated header and bibliographic information.


sonofmun:
I think `<listBibl xml:lang="ger">` probably either doesn't need an @xml:lang attribute or that it needs to be defined in `<langUsage>` as `<language ident="ger">German</language>`

sonofmun:

PR #1006, opened 2017-02-24
Update tlg2959.tlg008.opp-grc1__cts__.xml

AlisonBabeu:
Edited language and added in edition info


planatheisa:

sonofmun:
See discussion on #1001

sonofmun:
@PonteIneptique This `__cts__.xml` file is not passing Hook and I am not sure why. Hook says, "0/1 file(s) with descs", which I assume refers to the `<ti:description>` element on the `<edition>`. But it is there and looks like it is formatted correctly. What am I missing?

PonteIneptique:
You're missing xml:lang. `xml:lang` is mandatory.
Add "mul" if there is multiple language https://en.wikipedia.org/wiki/ISO_639-3#Non-language_codes

sonofmun:
Gotcha. I will just go ahead and do this to get it passing. I am going to recreate all of these shortly in any case.

PR #1007, opened 2017-02-24
Update info for tlg2959.tlg008.opp-grc1.xml

AlisonBabeu:
Fixed header and bibliographic info.


sonofmun:
See my comment about the @xml:lang attribute on #1005

sonofmun:

PR #1008, opened 2017-02-24
Update tlg2959.tlg010.opp-grc1__cts__.xml

AlisonBabeu:
Edited language and updated bibliographic info.


planatheisa:

sonofmun:
See discussion at #1001

PR #1009, opened 2017-02-24
Update tlg2959.tlg010.opp-grc1.xml

AlisonBabeu:
Added header info and updated bibliographic information.


sonofmun:
See my comments about @xml:lang attribute on #1005

sonofmun:

PR #1011, opened 2017-02-24
Update tlg2959.tlg011.opp-ger1.xml

AlisonBabeu:
Added in work title and author and updated bibliographic info


sonofmun:
I would remove the @xml:lang attributes on `<listBibl>` and the editor's `<name>` and then also remove Latin from `<langUsage>`.

sonofmun:
Change "de" to "deu". And could you add `@xml:lang="lat"` to `<titleStmt>/<title>`?

PR #1013, opened 2017-02-24
Update tlg2959.tlg012.opp-ger1.xml

AlisonBabeu:
Edited header and added bibliographic info.


sonofmun:
I would remove @xml:lang="lat" from `<listBibl>` and the editor's `<name>`. And if you could also add @xml:lang="de" to the `<div type="edition">` element, that would be great.

sonofmun:
Change "de" to "deu". And could you add `@xml:lang="lat"` to `<titleStmt>/<title>`?

PR #1017, opened 2017-02-28
Added tlg0062.tlg022

sonofmun:
Lucian Timon

PR #1016, opened 2017-02-28
Added tlg0062.tlg021

sonofmun:
Lucian Icaromenippus

PR #1018, opened 2017-02-28
Added tlg0062.tlg042 Lucian De morte Peregrini

sonofmun:


sonofmun:

planatheisa:
@sonofmun 
I have only just realized, that xml:lang should be "mul" for multiple languages. For the pull requests #1016 and #1017, which I just merged, they are still "lat". Do you want to change them now or in the inevitable overhaul of all the other files?

planatheisa:
Also, in sourceDesc/author has no language tag.

sonofmun:
Don't worry too much about the `__cts__.xml` files. I will recreate them in the proper format (probably) next week. But do definitely check the language tags in the teiHeaders closely (as you are doing).

planatheisa:
Quick question: Should I add `xml:lang="eng"` to the teiHeader? And would I need to add `<language ident="deu">German</language>` to langUsage for files with a German titel/author/... ?

sonofmun:
Yes, do this. Basically the language attribute is inherited by the descendants of a node, so since most of the teiHeader is in English, we should add xml:lang to the header and then mark the nodes that are not English. And if there is something that needs to be specifically marked as German, then we should define "deu" in langUsage and then use it. This is especially important for German (or other) translations that we have, i.e., to make xml:lang on `<div type="edition">` "deu" so that this gets marked as a translation in the metadata files instead of an edition.

PR #1021, opened 2017-02-28
Added tlg0062.tlg045.1st1K-grc1.xml Lucian De saltatione

sonofmun:


planatheisa:
No language attribute in sourceDesc/author.

planatheisa:
Here is also tlg0062.tlg044.1s1K-grc1...

PR #1029, opened 2017-03-02
Added 'Glossen und Scholien'

planatheisa:
Checking Hook.


planatheisa:
Looks good.

sonofmun:
`<monogr>/<title>` should have an `@xml:lang="deu"` attribute. In fact, I think I would put the `@xml:lang="deu"` on the `<monogr>` element, since really everything in there is German (Leipzig, Teubner). Then you can also take if off of the editor's name.

PR #1043, opened 2017-03-14
Added Porphyry Quaestionum Homericarum ad Illiadem fasc. I

planatheisa:
Checking Hook.


planatheisa:
Looks good.

sonofmun:
There is a `<div type="textpart" subtype="chapter" n="11">` that should, I think, be `@subtype="section"` instead.
And `<div type="textpart" subtype="chapter" n="19">` needs another closing div tag in front of it. Right now it appears that it is being counted as in the second level of citations (i.e., sections) instead of in the first, with the other chapters. I also think that this should be `@n="20"` instead of 19 and then the chapters afterwards should be renumbered.
Note: these were all observations from looking at the raw file on Github instead of at the file in Oxygen. If they are not correct, then let me know and I will pull this branch and take a look at the files locally with Oxygen.

planatheisa:
Actually, I missed to chapters. Corrected now.

PR #1044, opened 2017-03-14
Added Porphyry Quaestionum Homericarum ad Illiadem fasc. II

planatheisa:
Checking Hook.


planatheisa:
Looks good.

sonofmun:
One small change. The `<publicationStmt>/<date>` value should now be "2017". I haven't paid attention to this up to this point but we should use 2017 from now on.

planatheisa:
Changed it for both. Never actually looked at that, but three months in, it's probably a good idea to change it. :)

PR #1047, opened 2017-03-16
Update tlg1210.tlg002.1st1K-grc1.xml

AlisonBabeu:
Added in author and standardized sourceDesc.
I also shortened the edition title for consistency with the other work.

### DIFF
--- a/data/tlg1210/tlg002/tlg1210.tlg002.1st1K-grc1.xml
+++ b/data/tlg1210/tlg002/tlg1210.tlg002.1st1K-grc1.xml
@@ -1,12 +1,12 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <?xml-model href="http://www.stoa.org/epidoc/schema/latest/tei-epidoc.rng" schematypens="http://relaxng.org/ns/structure/1.0"?>
 <TEI xmlns="http://www.tei-c.org/ns/1.0">
-<teiHeader>
+<teiHeader xml:lang="eng">
 <fileDesc>
 <titleStmt>
-<title>De Ortibus et Occasibus</title>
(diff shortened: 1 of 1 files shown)


planatheisa:
@sonofmun @AlisonBabeu  I think there should be language attributes for `<editor>`, `<author>`, `<title>` and `<imprint>`.

AlisonBabeu:
Hi. I had been under the impression that we weren't adding language attributes to all these elements as it would take a fair amount of time, particularly in terms of all the files that have already been done. In this case, I didn't add the language tag to the editor's name, I just didn't remove it because it was already there and technically correct.  Should I remove it here for consistency rather than adding language tags to all the other elements?

sonofmun:
We need to add the language tags to any elements that do not fit the general language tag for the teiHeader, which should be English. We will then pull this stuff automatically into the metadata for these files. So if we change it here, it should flow to all other places, including (hopefully) the catalog.
In addition to what @planatheisa has mentioned, there should also be an `xml:lang="eng"` attribute on the teiHeader element itself.

AlisonBabeu:
Ah I'm seeing what I missed before that all of those other fields are in Latin, and in general we have left the fields in Latin rather than standardizing them to Teubner, Leipzig, etc. I think I need more coffee. I will add the Latin tags here and try to be consistent in doing so as well as I move forward today.

planatheisa:
That would also be the case for the other pull requests, as far as I have seen.

sonofmun:
Yes, that would be great @AlisonBabeu

AlisonBabeu:
I went back through all of them and think I caught any place where I needed to add in a language tag @sonofmun and @planatheisa

PR #1049, opened 2017-03-16
Update tlg1216.tlg001.opp-grc1.xml

AlisonBabeu:
I edited the data a bit here, namely taking the volume number out of the title, and changing the date to the publication date of the original volume 1912, since 1985 was just the reprinting date.

### DIFF
--- a/data/tlg1216/tlg001/tlg1216.tlg001.opp-grc1.xml
+++ b/data/tlg1216/tlg001/tlg1216.tlg001.opp-grc1.xml
@@ -4,7 +4,7 @@
   <teiHeader>
     <fileDesc>
       <titleStmt>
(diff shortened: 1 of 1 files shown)


sonofmun:
@AlisonBabeu Just a couple of notes for future work. There should always be a `xml:lang="eng"` attribute on the teiHeader. This will mark the default language of everything in the teiHeader as English. Then, as you did, anything that is not English in the teiHeader should have a language tag showing what it is. Then, finally, any language tags you add should be represented with a `<language>` element in profileDesc/langUsage, in the case of this file, `<language ident="lat">Latin</language>`. I will go through and add those things on these PRs, so this is basically only for future reference. And we can go back and recorrect the files that don't have this information in the future (sometime).

PR #1075, opened 2017-03-17
Update tlg1463.tlg001 __cts__.xml

AlisonBabeu:
Fixing language and adding edition data.


sonofmun:
It looks like the same edition, with the same URN, is listed twice in this file. Is there some reason for that or can the second one be deleted?

sonofmun:
Sorry, I see now. One is German and one is Greek. I didn't look closely enough.

Issue #1079, opened 2017-03-20
tlg1765.tlg004.1st1K-grc1 and tlg3118.tlg002.1st1K-grc1


gcelano:
These files contain each two div elements with @type="edition": this seems to be an error, in that the other texts do not have this double div.

PR #1083, opened 2017-03-21
Removed second <div type='edition'

sonofmun:
Also corrected language tags in the teiHeader
Issue #1079


sonofmun:
Passes Hook. Please check  @planatheisa

sonofmun:
Still passing Hook.

PR #1086, opened 2017-03-21
Added ending </div> to make XML well formed

Arithmeticus:
The split version of tlg4089.tlg003.xml was not well formed.


sonofmun:
@Arithmeticus Thanks a lot for this change. Just FYI, the texts that we actually serve to http://cts.dh.uni-leipzig.de/ are the ones located in the /data folder. It would be great if you wanted to take a look at those and do corrections. Let me know if you have any questions about how to get started.

Arithmeticus:
@sonofmun I'd be happy to be given direction. Just FYI, I'm using select files from this and the @PerseusDL collections to convert to @textalign TAN XML format.

PR #1088, opened 2017-03-21
Update tlg1487.tlg001__cts__.xml

AlisonBabeu:
Added in Greek work title and fixed edition metadata.


sonofmun:
Should we perhaps also add the Latin title as a child of ti:work?  I.e., `<ti:title xml:lang="lat">Carminum de actionum auspiciis reliquiae</ti:title>? Or is that not a generally accepted title for this work and thus a title that only applies to this edition?

AlisonBabeu:
I wasn't entirely certain about this, I have the Latin title in the catalog and assume I must of pulled it from the edition title, but I haven't found this title anywhere else and the TLG uses only the Greek.

sonofmun:
My feeling would be to add it since it is a possible title that someone might want to search under. My feeling is it is better to have more information than less.

PR #1089, opened 2017-03-21
Update tlg1487.tlg001.1st1K-grc1.xml

AlisonBabeu:
Added in author and work title to header as well as some language encoding.


sonofmun:
```
<teiHeader>
  <profileDesc>
    <langUsage>
```

PR #1090, opened 2017-03-21
Update tlg1487.tlg002__cts__.xml

AlisonBabeu:
Corrected work title and adding in language encoding information.


sonofmun:
I think the ti:label on the ti:edition here should be the Latin title given by the edition, i.e., "Carminum de actionum auspiciis reliquiae (epitome)". At least that is the way we have it on tlg1487.tlg001 `__cts__.xml`

sonofmun:
Also, the same question about adding the Latin ti:title element as on #1088

AlisonBabeu:
The reason I didn't add that here as well is I wasn't sure if that Latin name is actually used still or if it was just a convention of this edition.  I'm happy to add it here if you think its necessary @sonofmun

sonofmun:
I think it is a good idea to add it as a `<ti:title>` element alongside the Greek title on the `<ti:work>`. And since the title on the edition is  "Carminum de actionum auspiciis reliquiae (epitome)", I think that needs to be the `<ti:label>` on `<ti:work>/<ti:edition>` instead of the Greek title.

AlisonBabeu:
@sonofmun I think I've changed the cts file in the way you have requested, can you check this one out before I do the other one as well? Thanks!

sonofmun:
Looks good.

PR #1091, opened 2017-03-21
Update tlg1487.tlg002.1st1K-grc1.xml

AlisonBabeu:
Added in author info and language tagging.


sonofmun:
```
<teiHeader>
  <profileDesc>
    <langUsage>
```

PR #1096, opened 2017-03-21
Update tlg1551.tlg002.opp-grc1.xml

AlisonBabeu:
Added in author and work title, language tagging, series statement, etc.


sonofmun:
Added some more language tags, including xml:lang="grc" on the `<div type="edition">` element.

PR #1098, opened 2017-03-21
Update tlg1622.tlg001.1st1K-grc1.xml

AlisonBabeu:
Updated various information.


sonofmun:
```
<teiHeader>
  <profileDesc>
    <langUsage>
```

PR #1117, opened 2017-03-23
Update tlg3135.tlg001.opp-grc1.xml

AlisonBabeu:
Standardized metadata.
Changed file ID.

Also @sonofmun : Should we change the title of this work as it does not reflect the reality of the file? Only volume 1 of Dindorf's edition
of this work seems to be contained in this file, but I couldn't quite figure out how many books of the full work were within this file.


sonofmun:
@AlisonBabeu It looks like there are 18 books from the original work in this file (at least that is what the citation scheme says). So that means that the title with its additional (lib. 1‑12) is incorrect. How many books are in the complete work? It could be that we joined more than one volume together to get this many books.

sonofmun:
Wikipedia says that there are 18 books in the full work (https://en.wikipedia.org/wiki/Joannes_Zonaras#Written_works). So it appears that we have the whole work in this file.

AlisonBabeu:
Hi @sonofmun thanks for helping me figure this out and yes it looks like the entire work is included in the file. The TLG separates out Books 13-18 into tlg3135.tlg002, which we did not do, and the original metadata for this work is also incorrect in that it should have indicated this work was drawn from four volumes. I will fix both the work title and the edition metadata and submit a new pull request.

PR #1128, opened 2017-03-28
Corrected stoa0146d.stoa001

sonofmun:
Separated the Greek from the Latin

Inserted chapter divisions from the scanned edition

Corrected teiHeader


sonofmun:
Passes Hook. Please check @planatheisa or @annettegessner

PonteIneptique:
Was hast du mit dem Deutschen Text gemacht ?

sonofmun:
What do you mean?

PonteIneptique:
The whole intro has been removed it seems ?

sonofmun:
I removed anything that is not the text itself. That includes front and back matter.

sonofmun:
Perhaps I shouldn't have? I was still thinking in terms of CSEL where we have all of the original volumes also in the repo. Here we don't have that.

PonteIneptique:
I really think it's time we have a ti:commentary item for front...

PonteIneptique:
Let me think about this one. I might implement ti:commentary. It won't be CTS compliant but will be DTS compliant.

sonofmun:
That would be a good solution, I think. After all, if we say one file, one text, then we shouldn't be including the introduction to an edition with the edition itself. They are two different texts.

PonteIneptique:
I'll merge that but could you open an issue on CapiTainS.github.io repository ?

sonofmun:
On the github.io repository?

PonteIneptique:
https://github.com/Capitains/capitains.github.io/

sonofmun:
Got it. Will do.

PR #1143, opened 2017-03-29
Corrected tlg2042.tlg030

sonofmun:
Origens Commentary on Matthew

Separated the Greek from the Latin

Removed duplicate nodes

Corrected metadata file to contain both the edition and the translation

Updated the teiHeader


sonofmun:
Passing Hook.

PonteIneptique:
I have one issue : There is some time Greek critical apparatus in the Latin edition. Not sure it was meant to be. It might also mean you removed it from the Greek edition...

sonofmun:
I didn't remove any notes or critical apparatus from either version. To do that would have meant that I needed to look at every note to see if it refers to the Greek or Latin text. After all, sometimes a note on the Latin translation might refer to the original Greek, so I couldn't simply separate them according to what language they contain. So that means both versions should also contain the notes for the other version as well.

PonteIneptique:
Perfect.

PR #1144, opened 2017-03-29
Corrected tlg2703.tlg001 Anna Comnena Alexias

sonofmun:
Separated Greek from Latin

Removed duplicate nodes

Updated teiHeader

Updated metdata file


sonofmun:
All passing Hook.

PR #1154, opened 2017-04-03
Update tlg2018.tlg010__cts__.xml

AlisonBabeu:
Added in German title for that edition.
Edited language tags.
Added in title description


sonofmun:
The `<title>` tag on the first edition (urn:cts:greekLit:tlg2018.tlg010.opp-ger1) should be a `<label>` tag instead.

PR #1157, opened 2017-04-04
Update tlg2018.tlg011__cts__.xml

AlisonBabeu:
Updated language tagging and edition title data.


sonofmun:
Since tlg2018.tlg011.opp-lat1 is a translation of the Greek work (I assume), it should be in a ti:translation element instead of a ti:edition element.

AlisonBabeu:
I actually don't know the answer to this question, but if this is correct, I think this means we need to go back and alter a few other cts_.xml files because I'm pretty sure this isn't our only translation.

sonofmun:
That is true. But we are adding language detection in HookTest and so it soon should be clear where we need to go and change ti:edition to ti:translation. Don't worry about going back and looking for errors. But when you see them, go ahead and change them.

sonofmun:
One more note here about the ti:translation element. This element requires an xml:lang attribute with the value being the language of the translation. So we need to make sure to include this when we do translations.

PR #1160, opened 2017-04-04
Update tlg2018.tlg011.opp-lat1.xml

AlisonBabeu:
Added in author.
Added in language tags.
Edited bibliographic info.


sonofmun:
@AlisonBabeu One more part of doing the translations correctly is instead of having a `<div type="edition">` element that contains the text, it should be a `<div type="translation">`. Again, I think we will be able to find these texts pretty easily once we get the new Hook system up and running. I will go ahead and change it on this PR. This note was simply for future reference.

AlisonBabeu:
Consider it referenced @sonofmun!

PR #1194, opened 2017-04-11
Update tlg2036.tlg001.opp-grc1.xml

AlisonBabeu:
Edited language tagging
Added in author, work title and header
Edited bibliographic information and added in a series statement


planatheisa:
This doesn't pass: error: element "series" not allowed here [In (L78 C35)]

PR #1238, opened 2017-04-28
Update tlg2042.tlg084.opp-grc1.xml

AlisonBabeu:
Added in editor info
Added in language tagging
Updated bibliographic info

### DIFF
--- a/data/tlg2042/tlg084/tlg2042.tlg084.opp-grc1.xml
+++ b/data/tlg2042/tlg084/tlg2042.tlg084.opp-grc1.xml
@@ -3,8 +3,8 @@
 <TEI xmlns="http://www.tei-c.org/ns/1.0">
     <teiHeader xml:lang="eng">    
         <fileDesc>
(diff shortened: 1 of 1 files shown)


capitains-hook:
*[Hook UI build recap](http://ci.perseids.org/repo/OpenGreekAndLatin/First1KGreek/160)*

capitains-hook:
*[Hook UI build recap](http://ci.perseids.org/repo/OpenGreekAndLatin/First1KGreek/162)*

planatheisa:
Is it `xml:lang="deu"` or `xml:lang="ger"` ? I have seen both recently and I always thought "ger" to be the "right" one. @AlisonBabeu

sonofmun:
If you look here (http://www-01.sil.org/iso639-3/codes.asp?order=639_3&letter=d), the correct ISO 639-3 code for German is "deu". We should try to be consistent when using this, correcting it when we see a problem. Though, in the end, both will be correctly understood since there is no "ger" code in the 639-3 standard.

PR #1302, opened 2017-05-18
Update tlg2200.tlg00437.opp-grc1.xml

AlisonBabeu:
Added in editor
Added in language tagging
Edited bibliographic info

### DIFF
--- a/data/tlg2200/tlg00437/tlg2200.tlg00437.opp-grc1.xml
+++ b/data/tlg2200/tlg00437/tlg2200.tlg00437.opp-grc1.xml
@@ -1,12 +1,12 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <?xml-model href="http://www.stoa.org/epidoc/schema/latest/tei-epidoc.rng" schematypens="http://relaxng.org/ns/structure/1.0"?>
 <TEI xmlns="http://www.tei-c.org/ns/1.0">
-  <teiHeader>
-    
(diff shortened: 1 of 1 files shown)


capitains-hook:
*[Hook UI build recap](http://ci.perseids.org/repo/OpenGreekAndLatin/First1KGreek/673)*

planatheisa:
@AlisonBabeu 
Maybe there should be an `xml:lang="deu"` in `<imprint>`? There should definitly be `xml:lang="deu"` in <langUsage>.

AlisonBabeu:
Hi, I am not sure about ``` xml:lang="deu"``` in ```<imprint>``` because the publication city name Leipzig is not specifically German, nor is the publication year, were you thinking because of the publisher name B.G. Teubner?  What do you think @sonofmun?  I will definitely add in ```<xml:lang="deu">``` in langUsage and I missed the ```xml:lang="lat"``` on the initial title statement, thanks for catching that @planatheisa .

capitains-hook:
*[Hook UI build recap](http://ci.perseids.org/repo/OpenGreekAndLatin/First1KGreek/732)*

sonofmun:
I think if there is no difference between a name in English and a name in German, that we don't need the `xml:lang` on that element. Both Leipzig and Teubner are the same in English and German. When it comes to a name like, e.g., Göttingen, with the umlaut over the o, then things get a bit trickier. But here I think we don't need it.

PR #1322, opened 2017-05-19
tlg2371.001_cts_.xml and tlg2371.tlg001.opp-grc1.xml

AlisonBabeu:
cts_.xml file--Updated language tagging
Edited bibliographic info

tlg2371.tlg001.opp-grc1.xml
Added in language tagging
Edited bibliographic info

### DIFF
--- a/data/tlg2371/__cts__.xml
+++ b/data/tlg2371/__cts__.xml
@@ -1,3 +1,3 @@
 <ti:textgroup xmlns:ti="http://chs.harvard.edu/xmlns/cts" urn="urn:cts:greekLit:tlg2371">
-  <ti:groupname xml:lang="eng">Chronicon Paschale</ti:groupname>
+  <ti:groupname xml:lang="lat">Chronicon Paschale</ti:groupname>
 </ti:textgroup>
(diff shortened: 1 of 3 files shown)


planatheisa:
`<idno type="filename">` should be `tlg2371.tlg001.opp-grc1.xml` instead of `tlg2371.tlg001-opp-grc1.xml`.

AlisonBabeu:
HI @planatheisa so I fixed the file ID. I think this is all set now.

sonofmun:
@AlisonBabeu A couple more small changes: 
Could you change the `/publicationStmt/date` to 2017 instead of 2014?
Could you add `xml:lang="lat"` to `/biblStruct/series/title`?

planatheisa:
@sonofmun One more question: Is there any specific way how we handle `<biblScope unit="volume">`? In this file we have "vol. i." and "vol. 16", in other files it only gives the volume number (like "3" or "2.3"). Maybe we should find a definite way for this. (I'd suggest arabic numbers without any "vol").

sonofmun:
I agree with your suggestion. We should simply have an Arabic number, with decimal, comma, etc., for a multipart volume. Or, e.g, 4-6, for multiple volumes.

AlisonBabeu:
Hi @planatheisa and @sonofmun I'm glad someone finally brought up the issue of standards with the way we encode volume numbers. I hadn't been entirely certain and I've been fairly inconsistent myself I"m afraid. I will follow what is recommended above for all future headers.

sonofmun:
@AlisonBabeu Could you go ahead and change the two places with volume numbers that are in the header here?

AlisonBabeu:
Hey @sonofmun I think this is all set.

PR #1341, opened 2017-05-25
Update tlg2200.tlg00446.opp-grc1.xml

AlisonBabeu:
Added in editor
Added in language tagging
Edited bibliographic info

### DIFF
--- a/data/tlg2200/tlg00446/tlg2200.tlg00446.opp-grc1.xml
+++ b/data/tlg2200/tlg00446/tlg2200.tlg00446.opp-grc1.xml
@@ -54,7 +54,7 @@
                 <availability>
                     <licence target="https://creativecommons.org/licenses/by-sa/4.0/">Available under a Creative Commons Attribution-ShareAlike 4.0 International License</licence>
                 </availability>
-                <date>2014</date>
+                <date>2017</date>
                 <publisher>University of Leipzig</publisher>
                 <pubPlace>Germany</pubPlace>
             </publicationStmt>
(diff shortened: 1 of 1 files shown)


capitains-hook:
*[Hook UI build recap](http://ci.perseids.org/repo/OpenGreekAndLatin/First1KGreek/871)*

planatheisa:
@AlisonBabeu 
A few things:
Is the authors name speicifically Latin? In that case it should contains `xml:lang="lat"`.
`<publicationStmt>` gives the year 2014.

AlisonBabeu:
Hi again @planatheisa and @sonofmun. So in terms of the author's name Libanius, I haven't been encoding it as ```xml:lang="lat"``` in any of the other files, I think because I have only been encoding authors' names as Latin when there are specific separate English version of their names that are typically used, if that makes any sense. My other question is regarding the ```publicationStmt```, I had not been changing this date in any of the files as I had been going along, is that something I should start doing now? I had thought for some files published earlier in Open Greek and Latin, that the publication year would simply remain the same.

sonofmun:
@AlisonBabeu and @planatheisa I agree that we should only put `xml:lang="lat"` for an entry that would be different in Latin than in English, e.g., Aristotle or Aristoteles (I think). As far as the publication statement, I would put 2016 or 2017 since all of these files were first published in 100% CapiTainS compliant format within the context of 1st1K. To make things easier on us, anything that is already in the repo and we are just updating, let's mark as 2016 and any new things that we add now from the stuff we have gotten back from DDD we should mark 2017. What are your thoughts here?

planatheisa:
@sonofmun Sounds like a good idea. Maybe mention in the PR's if it's a new file, so we don't have to check every time?

capitains-hook:
*[Hook UI build recap](http://ci.perseids.org/repo/OpenGreekAndLatin/First1KGreek/939)*

capitains-hook:
*[Hook UI build recap](http://ci.perseids.org/repo/OpenGreekAndLatin/First1KGreek/950)*

AlisonBabeu:
O.k. I can do that as well for publication dates.

PR #1498, opened 2017-06-13
Update tlg3156.tlg001._cts.xml

AlisonBabeu:
Deleted the German title here because it was the title of the edition within which this Greek work 
was only part, not a translated German title for the work itself. have instead entered the Latin title of the work as given in the TLG and in spreadsheet.


capitains-hook:
*[Hook UI build recap](http://ci.perseids.org/repo/OpenGreekAndLatin/First1KGreek/1572)*

PR #1538, opened 2017-06-26
added URLs for figures/tables

brittanyhardy:


sonofmun:
Nice work here, @brittanyhardy

PR #1618, opened 2017-07-24
created new issue branch to resubmit

Pria20:


sonofmun:
It appears that in this issue, that tlg2001.tlg043.opp-grc1.xml was deleted instead of simply being renamed to tlg2001.tlg043.opp-lat1.xml

sonofmun:
Replaced by #1699

Issue #1568, opened 2017-07-12
tlg4024.tlg001.opp-grc2 also contains the Latin translation


AlisonBabeu:
[tlg4024.tlg001.opp-grc2](https://github.com/OpenGreekAndLatin/First1KGreek/blob/master/data/tlg4024/tlg001/tlg4024.tlg001.opp-grc2.xml)

The Historiae by Agathias Scholasticus, also contains the Latin translation of this work. I believe this means @sonofmun we need to split out the translation into the separate file. I did fix the header and other issues with the file via #1567

Pria20:
I can work on this one :)

PR #1619, opened 2017-07-28
(tlg4024.tlg001.opp-grc2/opp-lat1) separate translations

Pria20:


sonofmun:
It looks like the citation units were not copied from the Greek over to the Latin. So the text is there, but the `<div>`s for all of the chapters did not make it to the Latin version and I expect the ones that did are no longer in the Greek version. These need to be copied over to the Latin in the correct places and reinserted in the Greek where they are now missing.

lcerrato:
fix #1568

PR #1652, opened 2017-08-03
Update tlg5034__cts__.xml

AlisonBabeu:
Added in the English textgroup name that is used in the files.


capitains-hook:
*[Hook UI build recap](http://ci.perseids.org/repo/OpenGreekAndLatin/First1KGreek/2049)*

PR #1789, opened 2017-09-06
Update tlg0062.tlg002.1st1K-grc1.xml

AlisonBabeu:
So I've changed the language title to "mul" here but I'm wondering if we should have two titles instead:
Namely:
```xml
<title xml:lang="lat">Hippias</title>
<title xml:lang="eng">The Bath</title>
```
What do you think @sonofmun?

### DIFF
--- a/data/tlg0062/tlg002/tlg0062.tlg002.1st1K-grc1.xml
+++ b/data/tlg0062/tlg002/tlg0062.tlg002.1st1K-grc1.xml
@@ -4,7 +4,7 @@
 <teiHeader xml:lang="eng">
 <fileDesc>
 <titleStmt>
-<title xml:lang="lat">Hippias, or the Bath</title>
+<title xml:lang="mul">Hippias, or the Bath</title>
  <author>Lucian of Samosata</author>
 <editor>A. M. Harmon</editor>
  <funder>Harvard Library Arcadia Fund</funder>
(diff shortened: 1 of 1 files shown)


sonofmun:
I like that idea. I was not sure if those two were actually different titles or something like a title and a subtitle. If you look at the Greek of some of these works that have two parts to their titles, both parts actually exist in the Greek header. I am not sure whether this actually represents the traditional titular conventions for these works.

AlisonBabeu:
In this case I think its simply how the Loeb editor chose to translate the title, and from looking at the catalog record I have it as a separate English title, so I will go ahead and make that change then.

capitains-hook:
*[Hook UI build recap](http://ci.perseids.org/repo/OpenGreekAndLatin/First1KGreek/2658)*

sonofmun:
The Greek header for this work reads: ΙΠΠΙΑΣ Η ΒΑΛΑΝΕΙΟΝ, so both parts are there. Like I said, though, I don't know how that should relate to our titling practice.

capitains-hook:
*[Hook UI build recap](http://ci.perseids.org/repo/OpenGreekAndLatin/First1KGreek/2661)*

lmuellner:
Hi @AlisonBabeu and @sonofmun, if this is the title, shouldn’t there be an “or” between the two of them in the metadata?

AlisonBabeu:
Hmm. You raise a very good point @lmuellner, since "The Bath" is not a translation of Hippias and that is what the metadata now looks like. Perhaps @sonofmun we should go back to your original title, just keeping "mul" as the language tag, or namely
```xml
<title xml:lang="mul">Hippias, or the Bath</title>
```

sonofmun:
I think that is a good idea. Could you make those changes and do a new PR, @AlisonBabeu ?

PR #1793, opened 2017-09-06
Update tlg0062.tlg016.1st1K-grc1.xml

AlisonBabeu:
Updated header language tagging.
Also @sonofmun  I've noticed here that Lucian's name is in Latin (rather than in English)
as with the last files I just checked, I'm going to leave it, if that's all right, unless
you think that it should be the English version of the name in all the work files for consistency.


capitains-hook:
*[Hook UI build recap](http://ci.perseids.org/repo/OpenGreekAndLatin/First1KGreek/2669)*

sonofmun:
That's fine to leave it in Latin instead of English.

PR #1958, opened 2017-12-12
Update __cts__.xml

AlisonBabeu:
Edited Latin name variant and added in English name since there are a lot of Menanders
floating around! :)


capitains-hook:
*[Hook UI build recap](http://ci.perseids.org/repo/OpenGreekAndLatin/First1KGreek/3532)*

PR #1969, opened 2018-01-08
Greg's changes to tlg0011.tlg003

sonofmun:
I think the problem with the failing file is that there is no metadata file (i.e., `__cts__.xml` file) for either the author or the work. These will need to be added. @gregorycrane , if you want to do this, just grab an author level and a work level `__cts__.xml` file from one of the other authors and works and change it to match the metadata for the file. If you are not interested in doing this, let me know and I will make the fixes myself.


capitains-hook:
*[Hook UI build recap](http://ci.perseids.org/repo/OpenGreekAndLatin/First1KGreek/3593)*

data/tlg0011/tlg003/__cts__.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ti:work xmlns:ti="http://chs.harvard.edu/xmlns/cts" urn="urn:cts:greekLit:tlg0011.tlg003" xml:lang="grc" groupUrn="urn:cts:greekLit:tlg0011">
+   <ti:title xml:lang="eng">Ajax</ti:title>
+   <ti:edition xml:lang="grc" workUrn="urn:cts:greekLit:tlg0011.tlg003" urn="urn:cts:greekLit:tlg0011.tlg003.1st1K-grc1">
+      <ti:label xml:lang="lat">Sophoclis Aiax. Commentario perpetuo illustravit Chrst. Augustus Lobeck</ti:label>
```
planatheisa: This is different than what we usually do here. Should this be changed?

data/tlg0011/tlg003/tlg0011.tlg003.1st1K-grc1.xml
```diff
@@ -7,7 +7,7 @@
             <title type="work" n="Aj">Ajax</title>
```
planatheisa: There are a lot of little differences here to what we usually do. How much does this matter?

PR #1970, opened 2018-01-08
Update .travis.yml

sonofmun:
Removed the `ls -R` command. It is not necessary and produces log files that are too long for Travis.


capitains-hook:
*[Hook UI build recap](http://ci.perseids.org/repo/OpenGreekAndLatin/First1KGreek/3596)*

PR #1974, opened 2018-01-22
Separated the Latin from the Greek and created a new file for it.

ahanhardt:


data/tlg4024/tlg001/__cts__.xml
```diff
@@ -5,4 +5,9 @@
     <ti:description xml:lang="mul">Agathias Scholasticus, Historiae, Corpus Scriptorum Historiae Byzantinae, Vol 1,
     Niebuhr, Weber, 1828</ti:description>
   </ti:edition>
+  <ti:translation urn="urn:cts:greekLit:tlg4024.tlg001.opp-lat1" workUrn="urn:cts:greekLit:tlg4024.tlg001">
```
sonofmun: You need an `xml:lang="lat"` attribute on the `<translation>` element. Since a translation does not share the same language as the original, it needs to be specifically declared.

PR #2029, opened 2018-02-22
Update tlg0552_cts_.xml

AlisonBabeu:
Added in French name variant as used in most of the files.


capitains-hook:
*[Hook UI build recap](http://ci.perseids.org/repo/OpenGreekAndLatin/First1KGreek/3919)*

PR #2036, opened 2018-02-22
Update tlg0559_cts_.xml

AlisonBabeu:
Added in Latin variant (in one header) and in German name (because its in all the scans and looks cool! :P )

### DIFF
--- a/data/tlg0559/__cts__.xml
+++ b/data/tlg0559/__cts__.xml
@@ -1,6 +1,6 @@
 <ti:textgroup xmlns:ti="http://chs.harvard.edu/xmlns/cts" urn="urn:cts:greekLit:tlg0559">
   <ti:groupname xml:lang="eng">Hero of Alexandria</ti:groupname>
   <ti:groupname xml:lang="lat">Heronis Alexandrini</ti:groupname>
-  <ti:groupname xml:lang="deu">Herons von Alexandreia</ti:groupname>
+  <ti:groupname xml:lang="deu">Heron von Alexandreia</ti:groupname>
 </ti:textgroup>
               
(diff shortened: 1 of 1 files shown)


capitains-hook:
*[Hook UI build recap](http://ci.perseids.org/repo/OpenGreekAndLatin/First1KGreek/3933)*

data/tlg0559/__cts__.xml
```diff
@@ -1,3 +1,6 @@
 <ti:textgroup xmlns:ti="http://chs.harvard.edu/xmlns/cts" urn="urn:cts:greekLit:tlg0559">
   <ti:groupname xml:lang="eng">Hero of Alexandria</ti:groupname>
+  <ti:groupname xml:lang="lat">Heronis Alexandrini</ti:groupname>
+  <ti:groupname xml:lang="deu">Herons von Alexandreia</ti:groupname>
```
planatheisa: Where in the scan does it say that? Wikipedia gives the German name as "Heron von Alexandria".

sonofmun: Just bumping this comment @AlisonBabeu

AlisonBabeu: Whoops, sorry for missing this for so long, must have made a typo adding in the German name. The Latin name variant shows up in several of the `<biblstruct>` sections, like here: https://github.com/OpenGreekAndLatin/First1KGreek/blob/a718232fe587195ea5fd20821327a5a81c862b8f/data/tlg0559/tlg011/tlg0559.tlg011.1st1K-grc1.xml#L57 which is why I added it in.

I'll fix the German and you can do as you like with the Latin! :)

capitains-hook:
*[Hook UI build recap](http://ci.perseids.org/repo/OpenGreekAndLatin/First1KGreek/4130)*

PR #2056, opened 2018-03-01
Update tlg6014_cts.xml

AlisonBabeu:
Added in Latin variant appearing in the edition files


capitains-hook:
*[Hook UI build recap](http://ci.perseids.org/repo/OpenGreekAndLatin/First1KGreek/4027)*

PR #2074, opened 2018-03-26
tlg4090.tlg002 (incomplete)

Cami33:
I've tried to work on this file which was not complete, and added the metadata on the file level


Cami33:
Hi, how can I check the problems with this file?

lcerrato:
Hi @Cami33,

I will try to tackle a few of the issues I see here.

First, I think that the file name  tlg4090.tlg002.1st1K-grc1 (volume 1).xml is not allowed. Are looking to call this  tlg4090.tlg002.1st1K-grc1.xml? Is that correct? In any case, your CTS work file has one file name and your actual work file goes by another so that needs to be resolved. The parenthetical has to be removed or otherwise resolved and we have to make sure the file name is what it should be.

The next thing I see is, following on the file naming issue, that there is an incomplete `<div>` under the `<body>` tag in the text itself.
You have
`<body>`
   `<div type="bibliography">` 

whereas I expect the first `<div>` to declare three things: type, language, and the n attribute.
for example:
`<div type="edition" xml:lang="grc" n="urn:cts:greekLit:tlg4090.tlg002.1st1K-grc1">`

These two issues are why Hook reads `Naming convention failed` or `Naming conventions` as errors.

lcerrato:
Hi again, @Cami33 

I am now looking at the cts work file. There are some little inconsistencies to the way we work on these at Perseus, so I will ignore those and focus on what may be breaking the Hook tests.

Immediately, this jumps out at me
`<ti:work xmlns:ti="http://chs.harvard.edu/xmlns/cts" groupUrn="urn:cts:greekLit:tlg4090” xml:lang="grc" urn="urn:cts:greekLit:tlg4090.tlg002 (volume 1)”>`

The curly quotes here are not allowed.

I think there are more later on in the file, so check for those.

lcerrato:
see #2829

Issue #1682, opened 2017-08-10
tlg4034.tlg003.opp-grc1.xml contains two other works tlg4034.tlg004 and tlg4034.tlg005


AlisonBabeu:
While correcting the header for tlg4034.tlog003.opp-grc1.xml, "In libros de partibus animalium commentaria" by Michael Ephesius, @sonofmun I discovered that this text also contains the full text of two other works [tlg4034.tlg004](https://github.com/OpenGreekAndLatin/First1KGreek/blob/master/data/tlg4034/tlg003/tlg4034.tlg003.opp-grc1.xml#L4400) In Libros de animalium motione commentarium, (lines 4400-5608) and also [tlg4034.tlg005](https://github.com/OpenGreekAndLatin/First1KGreek/blob/master/data/tlg4034/tlg003/tlg4034.tlg003.opp-grc1.xml#L5610) In librum de animalium incessu commentarium (lines 5610 to 719). 

These will need to be separated out and made into two new work files.

ahanhardt:
If no one else is working on this issue, Lenny and I are going to tackle it with the CHS interns.

PR #2077, opened 2018-06-06
(tlg4034.tlg003.opp-grc1.xml) file contains additional works

jdeen33:
we separated tlg4034.004 and tlg4034.005 from tlg4034.003.


lcerrato:
fix #1682

PR #2082, opened 2018-06-15
formatted figures and layout of document 

charhoughton:
Worked on indentation, alignment, and fractions, as well as the figures. Not sure how well the XML will work as a web page, since I was kinda experimenting and fixing my code until it worked, but hopefully it'll be okay.


lcerrato:
@charhoughton 
That's a lot of work!

Issue #2086, opened 2018-06-26
Empty References


christmasunicor:
@lcerrato As I understand it, the latest version of [hooktest](https://github.com/Capitains/HookTest/releases) automatically checks for empty references. On one of my recent builds in Travis, I indeed receive the error "empty references" for about 4 files. (The log is [here, errors at lines 441ff.] (https://pastebin.com/GMcKD3Sq). Trouble is, I don't know what an "empty reference" is: could you elaborate? The documentation for HookTest is unfortunately sparse, and I haven't had any luck with the TEI guidelines either. :cold_sweat:

sonofmun:
Hi @christmasunicor, @lcerrato is exactly right here. This was just implemented last Friday in HookTest and the message should tell you at what level your empty reference is. I.e., if it says that you have 1 empty reference(s) at level 1, that means that if your citation scheme is book, chapter, section, that one of your `<div type="textpart" subtype="book">` elements has an empty `@n` attribute. This is something that is unique to HookTest and CapiTainS, since TEI is agnostic about even whether you have a citation scheme.

byeats:
Thanks for both of your comments: they cleared everything up. @lcerrato, I accidentally commented using my personal account, so @christmasunicor and I are the same person. See my pull request for these 4 fixed files. I know that we usually operate by "1 pull request, 1 file," but these files all correspond to this issue, so I ventured to keep them in one pull request. I can separate them, if necessary.

lcerrato:
@byeats 
Yes, one pull request per file allows for including the file name in the pull request. Our preferred format for an issue name or pull request name is  `(file name) issue description`so when multiple files are included, that doesn't work as well.
As noted in the pull request, this is fine for now — no need to break up the files.

PR #2087, opened 2018-06-28
Empty reference errors for multiple files 

byeats:
Fixed "empty reference" errors for 4 files - tlg0099.tlg001.1st1K-grc1.xml, tlg0527.tlg035.1st1K-grc1.xml, tlg0616.tlg001.1st1K-grc1.xml, and tlg0656.tlg001.1st1K-grc1.xml.


lcerrato:
@byeats 
Please remove any non-text files.
I do prefer one pull request per file but no need to change this issue for now.

I do not know why the travis file or the other file are in here, though.

lcerrato:
fix #2086

lcerrato:
Hi @PonteIneptique @sonofmun 
Should the `.travis.yml` be changed in this repo (as well as others)? Note this PR includes
`install HookTest>=1.1.6`

byeats:
@lcerrato Yes, those files shouldn't have been there. My .travis.yml was edited because I was doing some tests with HookTest on my local machine. Take a look at the recent changes and see if there is anything else I should fix. Thanks!

lcerrato:
@byeats 
Hi,
We may need to separate this pull request out for individual files for easier review. Let's see how it goes. There are a few different things I've spotted. Your empty divs actually point to a problem with the initial markup structure in tlg0616.tlg001.1st1K-grc1.xml These are all things that were in the file before you worked on it, so they would not be anything you would be on the lookout for.

I would keep ΘΗPΩN MIΔTΙAΔΟΥ. as a `<head>` as well as [APIΣOMΕENΕΣ MΕΣHNIOΣ.] and any other `<head>` tags that were removed in your most recent edits. The reason for this is that these are new sections.

It appears that this work is cited as book/chapter/section/line but the file reads book/section (no doubt how the file was initially structured via data entry). We should use book/chapter/section as our CTS containers (lines can remain milestones since this is prose.

Sections need to become chapters, which in turn, contain sections. Every chapter needs at least one section. And we want the refsDecl to reflect this added level of structure.

So I think that you need chapter 28 section 1 and chapter 28 section 2 because`<lb n="(2)"/>` refers to section 2 not another line break.  
https://archive.org/stream/polyaenistratege00polyuoft#page/n57/mode/2up

Note that this continues elsewhere: there are larger digits in the margins (1, 2, 3, 4, etc.) that designate sections. Chapter 29 has sections 1 and 2, as well, but these are tagged as line breaks — out of order and after the text. 

This isn't obvious because there aren't always digits for section 1. And if there is only one section (such as for book 1 chapter 22), we should still have a single section 1.

So there appear to be many structural issues with this work that require some revisiting. 

Also, if the chapter number is in the div, it need not appear again in the text. So here
`<div type="textpart" subtype="section" n="32">`
 ` <head>32. KINΕAΣ.</head>`
The  `<head>32. KINΕAΣ.</head>` contains an unnecessary 32. We've already captured that in the `<div>`

In looking at this, I noticed that some of the letters in the `<head> are not Greek. They are capital letters in English. It's not obvious depending on the browser, but you can see it in the raw file online. So let's make sure that all of the capitalized Greek is in fact Greek.

A helpful addition would have the resDecl  describe the structure of a text. So under the `<cRefPattern` tag  you should always have a `<p>` that describes what is going on, such as `<p>This pointer pattern extracts edition, parts, books, chapters and  sections.</p>` 
A good example of a three level refsDecl is here:
tlg0032.tlg007.perseus-grc2.xml
https://raw.githubusercontent.com/PerseusDL/canonical-greekLit/master/data/tlg0032/tlg007/tlg0032.tlg007.perseus-grc2.xml

I think this warrants a new issue with a checklist to cover all of these various issues for this particular work. You may either keep your changes local (stash the file or copy it outside of your local repo) and remove the file from this pull request or we can go ahead with this version and then create a new dedicated issue to address this work.

-Lisa

sonofmun:
`HookTest >= 1.1.6` should be fine.

lcerrato:
Thanks, @sonofmun. I wasn't sure how we should manage this within the repos.

byeats:
@lcerrato Well, goodness, this is a proper mess, isn't it? Thank you for going through the diff so carefully. Can we stick with this version and open a separate issue? That will allow HookTest to pass the other interns' pull requests to go through, in the mean time. Thank you!
-Ben

lcerrato:
see https://github.com/OpenGreekAndLatin/First1KGreek/commit/9fd687bb7fec2b0d29a54c775cb108a03f9c9d74

PR #2135, opened 2019-05-14
Update tlg0615.tlg001__cts__.xml

AlisonBabeu:
Fixed languages and added in some edition metadata


capitains-hook:
*[Hook UI build recap](http://ci.perseids.org/repo/OpenGreekAndLatin/First1KGreek/5016)*

PR #2158, opened 2019-10-22
Update tlg0614.tlg001.1st1K-grc1.xml

AlisonBabeu:
Corrected edition metadata. It was correct in the _cts file but not in the TEI XML file.


lmuellner:
Hi Alison, is Travis working now? Or are you doing this by hand? Best, Lenny

AlisonBabeu:
Hi @lmuellner. Travis right now is still not working unfortunately, but I normally do minor fixes like this by hand, since I discovered the metadata in the header was incorrect when I was checking on a different text.

sonofmun:
Hi All,
The problem with HookTest is that it is using the newest version of MyCapytains, with which it is not yet compatible. I have proposed a small change to HookTest and that should get your tests running again once we merge that PR.

lcerrato:
@sonofmun @AlisonBabeu 
And much rejoicing was had! Thank you!

sonofmun:
If the Hook Tests break again, just head over there (https://github.com/Capitains/HookTest) and open an issue. I only realized there was a problem because of the message that @lmuellner posted on this thread.

lcerrato:
@sonofmun 
Thanks. I was waiting on a second opinion from others (Thomas, James, etc.) as I was not sure if it was something I/we had checked in or not or if it was simply something else.

PR #2172, opened 2020-02-05
Update tlg0086.tlg001.1st1K-grc1.xml

AlisonBabeu:
The link was incorrect so I added in an Internet Archive link.


AlisonBabeu:
I added in an incorrect link too, my bad. I'm going to close this pull request add a correct link.

PR #2187, opened 2020-05-12
update gitignore so it ignores TEItoCEX and catalog.json files

ThomasK81:


ThomasK81:
@lcerrato or @AlisonBabeu  could you approve this please. It's just a minor change in the .gitignore

PR #2228, opened 2020-07-17
(tlg0609.tlg001.1st1K-grc1.xml) new files

### LINKED ISSUE #2214: (tlg0609.tlg001.1st1K-grc1.xml) new edition
Edited edition and work CTS file
[tlg0609.tlg001.zip](https://github.com/OpenGreekAndLatin/First1KGreek/files/4875359/tlg0609.tlg001.zip)
- [x] fix header

- [x] add refsDecl

- [x] fix body tags

- [x] create metadata files

- [x] check catalog and CITE tables

- [x] hyphenation and specialized tags; languages

- [x] special issues found on Trello card?

- [x] Update pick list and Trello card when complete

mkonieczny9805:
fix #2214


lcerrato:
@mkonieczny9805 
These appear to be the Tryphon files not the tlg0609 files. I'm going to close to avoid conflicts, but maybe you have the files on another branch?

mkonieczny9805:
I thought tlg0609 is Tryphon? (This is a different Tryphon from Tryphon Grammaticus)

lcerrato:
@mkonieczny9805 
Ok, the its workgroup is Tryphon Grammaticus, so that's got me confused. What does that mean for tlg1763?

lcerrato:
Ok, TLG has them as Tryphon I Grammaticus (609) and Tryphon II Grammaticus (1763). But we don't have that in the catalog, so I guess I have to fix that. I did not notice it previously because only one is cataloged so there are no I/II there.

mkonieczny9805:
@lcerrato Yeah I was just going by what's on the picklist, where tlg1763 is Tryphon Grammaticus, and tlg0609 is plain old Tryphon.

lcerrato:
@mkonieczny9805 
Yes, this has not been cataloged, so that is part of the confusion,  and, for future reference, I would not use the pick list for any final data — not for cts metadata or file headers, as most of the time that is shorthand. It's not catalog quality.
Also, the 1763 files are also attached here, which will create a conflict.

mkonieczny9805:
It's possible I made a mistake in switching between branches and included the tlg1763 files in this commit. Sorry for that.

Where should I get data from in the future? If something hasn't been catalogued yet and the picklist is shorthand, then it seems like the only authoritative resource would be TLG.

lcerrato:
@mkonieczny9805 
If there is a conflict between the picklist and the TLG or the TLG is obscure, just check with Alison first. VIAF or CERL are also available (sometimes Wikipedia or Loeb), and there is data that has been cataloged but is not visible in the catalog.  
I always double check with Alison if I've never heard of the author.

PR #2296, opened 2020-12-11
(tlg0317.tlg001.1st1K-eng1) new edition

### LINKED ISSUE #2271: (tlg0317.tlg001.1st1K-eng1) new edition
[urn.cts.greekLit.tlg0317.tlg001.1st1K-grc1.zip](https://github.com/OpenGreekAndLatin/First1KGreek/files/5373917/urn.cts.greekLit.tlg0317.tlg001.1st1K-grc1.zip)

- [x] fix header

- [x] add refsDecl

- [x] fix body tags

- [x] add first page link as milestone (if not possible to link in header)

- [x] create metadata files (if needed)

- [ ] check catalog and CITE tables

- [x] hyphenation and specialized tags such as q, add, del, sic; languages; digits in text

- [x] make sure there are spaces before/after page break tags and merge split words broken by pb tags

- [x] check for period at end of section (to avoid missing word drop out)

- [x] line breaks and special indentation 

- [x] special issues found on Trello card?

- [ ] Update pick list and Trello card when complete

charhoughton:
fix #2271


charhoughton:
fix #2271

lcerrato:
@charhoughton 
This is a little confusing — I think I'm seeing new metadata and the wrong file? Is this eng2? Our descriptions for both this PR and the issue itself are wrong.
I know I have an older file that was only partly done so I have to backtrack to figure this out.

charhoughton:
@lcerrato yes, this is eng2. My reasoning behind it was that I wan't sure if the Mendelssohn Greek also had a translation, and didn't want to create confusion further down the road (alas). I edited the metadata for the text to reflect that. In the original issue, it was described as a new edition, and I kept that phrasing for consistency. Let me know if that clears anything up

lcerrato:
@charhoughton 
Some of these changes are reflected in the first file.

So for instance, we have been using `<p rend="center">` this may be deprecated but if so I will have to make a global change across several sets of data. So I'd rather be consistent with existing practice. Also, be careful of things like labels (where the lines are numbered within sections — lines are strictly poetry so many of these are just paragraphs with labels rather than lines with verse numbers), and we need to remove all entities so nothing should have an ampersand there. 
Also for reading convenience, it's good to eliminate indentations.

I will take a look at these, but I see some issues that I fixed in the other batch, so it would be good to keep an eye on these moving forward.

It's difficult for me to push changes to another pull request and once I rename a file, the changes are not easily viewed, but I will try to do things so that you have an example. It may not be possible in this format.

lcerrato:
@charhoughton 
Ok, we don't set aside IDs to align with versions. So, grc1 and eng1 have no implied relationship. Most of the time, we are never going to have translations and editions from the same volume, in fact. (The rare exception would be a Loeb). 

Also, the ID assignments don't actually track things like which edition of the Greek the translator used. 

We just assign IDs as we go, so if this is our first translation, we would use eng1.  The only reason to use a different id would be to avoid duplication.

In general, I refer these questions to @AlisonBabeu before we commit. Unfortunately, there is no master system for this that generated these for us, so that we know which digits to use. This is why it is part of the issue checklist.

charhoughton:
@lcerrato thank you so much. I'll make sure to check these in the future.

charhoughton:
I'm actually not sure how to access the CITE tables, but I did check the catalog before pushing.

lcerrato:
@charhoughton 

Sorry: this differs from the Plutarch/Perseus workflow here because this is new stuff.  This is something I have to address in Trello because it's not really well documented. Basically, the Trello board workflow assumes that I would always be the one editing the files and I would check the file name as soon as it was moved over here. So the workflow is faulty in this case. :( It's nothing that would be obvious to anyone else. Unfortunately, I am not always going to be able to catch this sort of thing upfront.

It should have been flagged as using a placeholder id. (I think there is a checkbox in Trello where the proofreader is supposed to verify that they asked Alison or myself about the file URN/id before they continue with the work?)

The checklist here on Github is shorthand for what needs to be done, which is basically make sure we have the right file name (URN/ID): we first check the catalog, and if we have a match, we can bump some ids in the CITE tables (this is not publicly accessible), and then if neither of those have the information, we have to check the catalog pending information and/or get confirmation from Alison on the correct ID. So you won't be able to see all of the pending/on-going work in the catalog itself. 

You can't see the CITE tables without access (but it would not make a difference in this case). The catalog itself does not reflect current work as all of the records are not online. 

With the Plutarch, we were just making new editions of existing records but we don't have that here.

I will reemphasize this in Trello because it causes lots of confusion.

Sorry about that.

charhoughton:
@lcerrato, thanks for the explanation, that clears a lot up. I'm still used to the Perseus workflow. If you want me to make some of the edits and submit another pull request, I can.

AlisonBabeu:
Sorry to chime in late here. Thank you @lcerrato  for providing such a clear overview of the ID assignment process such as it is and all of its complications.  The numbering of translations and how they do or don't align with original editions has always been challenging in the creation of metadata for both Perseus and OGL.  Please feel free @charhoughton to tag me on any issues if there is a problem or question regarding an ID for a file as well!

lcerrato:
@AlisonBabeu 
So is this the correct URN for this file?

AlisonBabeu:
hi @lcerrato yes, to the best of my knowledge this should be eng1 as it is the first English translation of this work in any of my metadata.

lcerrato:
fixed via #2299

PR #2329, opened 2021-03-23
Add Scaife Viewer config file

jacobwegner:
Used to map attribution information from https://docs.google.com/spreadsheets/d/1hn5BrO7vEGugon_G5_AAahNfwSaWyG_V5usxz6GRrL0/edit#gid=522035422

When we extract attribution data from this repository for display on scaife.perseus.org:
- The `publisher` is extracted from `publicationStmt`
- Each additional `respStmt` is extracted, ending up with an attribution record with a person, organization or person + organization, and a role
- Scaife Viewer preserves the order of each `respStmt` as listed in the file.

The config file performs overrides as follows:

- `attributions.substitutions` allows us to map the spreadsheet corrections on top of the original files. If / when edits to the original files are made to update the headers, the mapping won't find a match, and we'll rely directly on the file.
- `attributions.promoted` lists `resp` values that should be sorted to the top of the attributions list for a given file (e.g. "Publisher" or "Proofreader")

### DIFF
--- a/.scaife-viewer.yml
+++ b/.scaife-viewer.yml
@@ -1,0 +1,549 @@
+attributions:
+  promoted:
+  - Publisher
+  - Additional proofreading and CTS conversion
+  - annotation
+  - Corrected and encoded the text
+  - CTS conversion
+  - Keyboarding
+  - proofreader
+  - Proofreading
+  - Proofreading and CTS conversion
+  - source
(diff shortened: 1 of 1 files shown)


jacobwegner:
@lcerrato Here is the pull request adding the configuration file.

PR #2421, opened 2021-06-21
(tlg0656.tlg002.1st1K-grc1.xml) new edition

### LINKED ISSUE #2370: urn:cts:greekLit:tlg0656.tlg002.1st1K-grc1.xml (new edition)
[b21459162_0003 (1).zip](https://github.com/OpenGreekAndLatin/First1KGreek/files/6380375/b21459162_0003.1.zip)

charhoughton:
fix #2370


charhoughton:
#2370

lcerrato (changes requested):
The structure in TLG is book, chapter, section.
The first chapter in Book one is "Pr" with the first chapter in Book two as "Arg"

data/tlg0656/tlg002/__cts__.xml
```diff
@@ -0,0 +1,7 @@
+<ti:work xmlns:ti="http://chs.harvard.edu/xmlns/cts" groupUrn="urn:cts:greekLit:tlg0656" xml:lang="grc" urn="urn:cts:greekLit:tlg0656.tlg002">
+    <ti:title xml:lang="lat">Euporista vel De simplicibus medicinis</ti:title>
+  <ti:edition urn="urn:cts:greekLit:tlg0656.tlg002.1st1K-grc1" workUrn="urn:cts:greekLit:tlg0656.tlg002">
+      <ti:label xml:lang="lat">De materia medica</ti:label>
```
lcerrato, line 4: What is the basis for this label? I see Περί απλών φαρμάκων on the title page here.

data/tlg0656/tlg002/__cts__.xml
```diff
@@ -0,0 +1,7 @@
...
+      <ti:description xml:lang="mul">Dioscurides Pedianus, Euporista vel De simplicibus medicinis, Wellmann, Weidmann, 1907-1914</ti:description>
+  </ti:edition>
```
lcerrato, line 6: The description should read:
Dioscorides Pedianus. Pedanii Dioscuridis Anazarbei de materia medica libri quinque, Vol. 1. Wellmann, Max, editor. Berlin: Weidmann, 1907.

[short author] [full book title] [last, first, editor/translator] [city: publisher, year].

lcerrato:
@charhoughton 
It looks like a space crept into your cts file name?
__cts__ .xml versus __cts__.xml

lcerrato (changes requested):

data/tlg0656/tlg002/__cts__.xml
lcerrato: Also, always check the info in the book against the spreadsheet as I noticed some differences in the volume and publication year here. That's not uncommon as the spreadsheet edition may not always be the exact one that ended up in Lace.

charhoughton: @lcerrato just to make sure: do we want the title from the picklist or the title page of the text?

data/tlg0656/tlg002/__cts__.xml
charhoughton: whoops, holdover from the metadata I was basing it off

data/tlg0656/tlg002/__cts__.xml
lcerrato: @charhoughton
This is a good question!

The metadata author and work title on the Google doc pick list are usually derived from the catalog or added by Alison, with some exceptions, so it would be rare to deviate from those as they most likely reflect our canonical names. If something looks off (say, the TLG is different), we can always just loop Alison in for a question. Keep in mind, she does see the data at the end of this process, because I create a new catalog_pending issue for all new editions. So that's another set of eyes.

(Plus, we will soon have a means for readers to report variant names and work titles, too, so the crowdsourcing will bring in better data and more corrections. So whatever inconsistencies or errors that may slip by, we're going to offer the users the chance to offer a corrections and additions.)

Much of the early data is inconsistent, so we're just doing the best we can.

When you get to the `<ti:label>` in the cts file, you want to reflect the print title on the page if it varies from the canonical title. This usually matches the `<title>` in the edition or translation file, as well, but we haven't been as strict there. Just think of it as you are digitizing the print, so you want to reflect what the print says, but when we put this in a table of contents, we're going to use the best "canonical" title, which is where the cts file comes in. It gets super messy, with author names and variations, but that's the basic rule of thumb for work titles.

To be honest, it doesn't matter much in the end. It's just in this case I didn't know where the label came from. (If it was just a cut and paste thing, no big deal. I've probably done this 100x times more than anyone else....)

PR #2423, opened 2021-06-22
new edition (tlg0057.tlg037.1st1K-grc1)

### LINKED ISSUE #2392: urn:cts:greekLit:tlg0057.tlg037.1st1K-grc1 [New Edition]
[urn.cts.greekLit.tlg0057.tlg037.1st1K-grc1.zip](https://github.com/OpenGreekAndLatin/First1KGreek/files/6543022/urn.cts.greekLit.tlg0057.tlg037.1st1K-grc1.zip)

Editing Trello Card: [https://trello.com/c/yJzga1Ji](url)
Publishing Trello Card: https://trello.com/c/ijiay913

charhoughton:
fix #2392


lcerrato:
@charhoughton 
`<p rend="merge">` seems to have crept into this file. (It was also shown as "μεργε" in two spots). It's extra work and not needed. 
The most significant question I have is that the TLG uses Kuhn volume/pages and doesn't have book/chapter/section. The problem we have here is that Kuhn volume and pages aren't preserved in any way, and if the TLG is using them, then we probably want them in the file. Were they in the Lace export?

charhoughton:
Hi @lcerrato, I'm not sure if the Kuhn divisions were in the lace export, but I don't think they were--I didn't do much reformatting on the divs. 

Thanks for letting me know that <p rend="merge"> is redundant. I'll take care of those (and the few in Greek...oops) next week. I can also reformat it so it agrees with TLG if we want

charhoughton:
@lcerrato would you prefer I add the Kuhn pages as milestones or restructure the text so they are used instead of book/chapter/section for the divs?

lcerrato:
@charhoughton 
If you add the Kühn pages as milestones, that would be great. It's not necessary, though.
If you don't want to prioritize, just create a new issue for this after we push this edition.
`<milestone unit="page" resp="Kuehn" n="6.453"/>`

charhoughton:
@lcerrato I can take care of it, it shouldn't be too difficult. The Kühn pages are in the CMG text, so I just have to page through and add them

lcerrato:
@charhoughton 
Just looking at this, were the Kühn pages put in the middle of any words? 

Such as `ὀμι|<milestone unit="page" resp="Kuehn" n="6.460"/>λήϲῃ`
`ὑποχ<milestone unit="page" resp="Kuehn" n="6.466"/>ώρηϲίν`

We should either put these before or after complete words, and always have spacing. Otherwise this will break any morphology tools and word searching.

Thanks,

lcerrato (approved):

charhoughton:
Yes, a lot of the Kühn pages were in the middle of words. Thank you for taking care of that

PR #2446, opened 2021-07-15
(urn:cts:greekLit:tlg0081.tlg013.1st1K-grc1.xml) new edition

### LINKED ISSUE #2393: (urn:cts:greekLit:tlg0081.tlg013.1st1K-grc1) new edition
urn:cts:greekLit:tlg0081.tlg013.1st1K-grc1

To access XML files and to see more information, see [Trello Card](https://trello.com/c/Hit8LAWz/14-edit-uc1c020256950-1583112705-dionysius-of-h-opuscula)

charhoughton:
fix #2393


data/tlg0081/tlg013/__cts__.xml
```diff
@@ -0,0 +1,7 @@
+<ti:work xmlns:ti="http://chs.harvard.edu/xmlns/cts" groupUrn="urn:cts:greekLit:tlg0081" xml:lang="grc" urn="urn:cts:greekLit:tlg0081.tlg013">
+    <ti:title xml:lang="lat">De compositione verborum (epitome)</ti:title>
+    <ti:edition urn="urn:cts:greekLit:tlg0081.tlg013.1st1K-grc1" workUrn="urn:cts:greekLit:tlg0081.tlg013">
+        <ti:label xml:lang="grc">Διονυσιου Ἁλικαρασεως τῶν περὶ μιμησεως τα σωιζομενα</ti:label>
+        <ti:description xml:lang="mul">Dionysius of Halicarnassus. Διονυσιου Ἁλικαρασεως τοῦ περὶ σινθεσεως ὀνοματων ἐπιτομη. Opuscula 2.1. Usener, Hermann and Radermacher, Ludwig, editors. Leipzig: Teubner, 1904.</ti:description>
```
lcerrato, line 5: @charhoughton I think this slipped past me with the other file, but if we use the Greek, we need the correct accentuation. This is often hard to deduce as titles are usually all caps. 

For tlg0081.tlg014 I changed the `<label>` back to Latin as the Greek did not appear on the book page (it was more of a header than a real title). 

But I see that both this description and the description for the other file aren't accented Greek. This can usually be derived from other sources or deduced via searching.

data/tlg0081/tlg013/tlg0081.tlg013.1st1K-grc1.xml
```diff
@@ -0,0 +1,1192 @@
...
+        <licence target="https://creativecommons.org/licenses/by-sa/4.0/">Available
+        under a Creative Commons Attribution-ShareAlike 4.0 International
+        License</licence>
+      </availability>
+      <date when="2021-05-25" />
+    </publicationStmt>
+    <sourceDesc>
+      <biblStruct>
+        <monogr>
+          <editor>Usener and Radermacher</editor>
```
lcerrato, line 128: Please add each full name on separate lines. See also tlg0081.tlg014 as that file is missing the same info.

data/tlg0081/tlg013/tlg0081.tlg013.1st1K-grc1.xml
```diff
@@ -0,0 +1,1192 @@
...
+          <title xml:lang="lat">Dionysii Halicarnasei Opuscula 2.1</title>
+          <imprint>
+            <pubPlace>Leipzig</pubPlace>
+            <publisher>Teubner</publisher>
+            <date>1904</date>
+          </imprint>
+        </monogr>
```
lcerrato, line 135: Please note the volume in the biblScope tag. Opuscula can be noted as a series.
Note that this is inconsistent with tlg0081.tlg014, so it might be good to revisit both files to come up with a uniform description.
I see that tlg0081.tlg014 is missing editor names.
We should have two lines:
```
<editor>first, last</editor>
<editor>first, last</editor>
```
This is how tlg0081.tlg014 reads and —while it needs the editor correction— use this as the model for this volume, too. (Note the series and volume info).
```
<biblStruct>
        <monogr>
          <title xml:lang="lat">De imitatione (fragmenta)</title>
          <author>Dionysius of Halicarnassus</author>
          <editor>Usener and Radermacher</editor>
          <imprint>
            <pubPlace>Leipzig</pubPlace>
            <publisher>Teubner</publisher>
            <date>1904</date>
          </imprint>
        </monogr>
        <series>
          <title xml:lang="lat">Opuscula</title>
          <biblScope unit="volume">2.1</biblScope>
        </series>
```

data/tlg0081/tlg013/tlg0081.tlg013.1st1K-grc1.xml
```diff
@@ -0,0 +1,1192 @@
...
+        <ref target="https://hdl.handle.net/2027/uc1.c020256950?urlappend=%3Bseq=155">HthiTrust</ref>
```
lcerrato, line 136: typo in HathiTrust

data/tlg0081/tlg013/tlg0081.tlg013.1st1K-grc1.xml
```diff
@@ -0,0 +1,1192 @@
...
+      <correction>
+        <p>Lace v. 0.6.18 copyright 2013-2020, Bruce Robertson, Dept. of Classics,
+        Mount Allison University. Developed through the support of the 
+        <ref target="https://nbif.ca/en">NBIF</ref>.</p>
+      </correction>
+    </editorialDecl>
+    <p>The following text is encoded in accordance with EpiDoc standards and with the
+    CTS/CITE Architecture</p>
+    <refsDecl n="CTS">
+      <cRefPattern n="book" matchPattern="(\w+)" replacementPattern="#xpath(/tei:TEI/tei:text/tei:body/tei:div/tei:div[@n='$1'])">
```
lcerrato, line 151: Is this correct? we have a cref pattern n="book" and a descriptor indicating "section" — your pattern needs to match the container description in the document

lcerrato (changes requested):

data/tlg0081/tlg013/__cts__.xml
lcerrato: The volume title is simlply Dionysii Halicarnasei Opusula Vol. 2.1
I see no need for any Greek in this description. (Don't use the inside plate, just what is here:
https://babel.hathitrust.org/cgi/pt?id=uc1.c020256950&view=2up&seq=8&skin=2021

So the label should be the same as the title (Latin), and your description needs to change as follows:

`<ti:description xml:lang="mul">Dionysius of Halicarnassus. Dionysii Halicarnasei Opusula Vol. 2.1 Usener, Hermann and Radermacher, Ludwig, editors. Leipzig: Teubner, 1904.</ti:description>`

Please also revisit tlg0081.tlg014 when this is done.

charhoughton: @lcerrato okay, I will.

 I've been a little confused about what needs to be in the title vs. the label vs. the description--the title should be what is in the picklist, the label should be the title as in appears in the book (if different from the picklist), and does the description need either of those? I know this needs to be accurate, so I want to be precise and hopefully sort it out for the future.

lcerrato: Yes, the problem is that we had no standards set up for a while and now we are struggling to be consistent.

The title is usually the canonical title, which is in the Catalog (or absent that)or  the pick list. Sometimes we may not find these to be good, but that's where you check with Alison.

The label is what is in the book. So for Greek, this is often Greek. (It's not a big deal if you keep the canonical name here). The label should match the title in the file itself.

The description is what should be thought of as a bibliographic short reference.
Common Author Name. Book Title, Vols (if applicable). Editor/Translator name/role. PubCity: Publisher, Year. 

The book titles can get long so we want something that is clear but not too long. Usually the first line or two on the bookplate.

lcerrato:
pinging @AlisonBabeu for reference

data/tlg0081/tlg013/__cts__.xml
charhoughton: @lcerrato okay, that definitely helps to clear things up. Do we then want the title and book title in the description to match the Catalog/picklist, and then the title in the text file would match the label, and come from the book? Or just make it easier and use the catalog title everywhere? Many thanks for everything

PR #2465, opened 2021-08-18
(tlg0575)

AlisonBabeu:
tlg0575.tlg001 -cts and xml
tlg0575.tlg002-cits and xml

Corrected editor name from Schenkle to Schenkl.


lcerrato:
@AlisonBabeu 
I think something is wrong here as these files look like they are in an earlier pull request (the one I commented on earlier). This one is also very large. Could it be that your branches got combined? 

We may need to revisit these as smaller PRs, as otherwise the work on the earlier branch is going to create a conflict when we go to merge this one.

So I would take a look and see if there is overlap here before proceeding.

PR #2523, opened 2021-09-14
(tlg2200).-_cts.xml and header updates

AlisonBabeu:
Have begun cataloging First1K expressions of LIbanius, updating cts_xml files and bibliographic descriptions to current standards.


data/tlg2200/tlg001/__cts__.xml
```diff
@@ -1,7 +1,7 @@
 <ti:work xmlns:ti="http://chs.harvard.edu/xmlns/cts" groupUrn="urn:cts:greekLit:tlg2200" xml:lang="grc" urn="urn:cts:greekLit:tlg2200.tlg001">
   <ti:title xml:lang="lat">Epistulae 1-1544</ti:title>
```
lcerrato, line 2: @AlisonBabeu 
I don't understand why the title would be 1-1544 when the label is 1-839. (plus I see only 831 letters in here).
Wouldn't the title be better presented as `<ti:title xml:lang="lat">Epistulae</ti:title>`?

data/tlg2200/tlg00401/__cts__.xml
```diff
@@ -1,7 +1,7 @@
-<ti:work xmlns:ti="http://chs.harvard.edu/xmlns/cts" groupUrn="urn:cts:greekLit:tlg2200" xml:lang="grc" urn="urn:cts:greekLit:tlg2200.tlg00401">
+<ti:work xmlns:ti="http://chs.harvardx.edu/xmlns/cts" groupUrn="urn:cts:greekLit:tlg2200" xml:lang="grc" urn="urn:cts:greekLit:tlg2200.tlg00401">
```
lcerrato, line 1: reads **harvardx**

lcerrato: This may be what's breaking the hook test but am unsure — it's something in this file.

lcerrato (changes requested):

data/tlg2200/tlg001/tlg2200.tlg001.opp-grc1.xml
```diff
@@ -65,12 +65,11 @@
...
                                 <persName>
-                                    <name xml:lang="lat">Richardus Foerster</name>
-                                    <addName xml:lang="deu">Richard Foerster</addName>
+                                    <name>Richard Foerster</name>
                                 </persName>
                             </editor>                           
-                            <title type="main">Libanii Opera</title>
-                            <title type="subordinate">Epistulae 1-839</title>
+                            <title type="main" xml:lang="lat">Libanii Opera</title>
+                            <title type="subordinate" xml:lang="lat">Epistulae 1-839</title>
```
lcerrato, line 72: There's an inconsistency with `<title type="subordinate"...>` versus `<title type="sub"...>`. You mainly use the latter.

data/tlg2200/tlg001/__cts__.xml
AlisonBabeu: I think I left that there because that's the total number of letters and wanted to remind myself that all of them weren't in the file.  Epistulae would be a much better title.

data/tlg2200/tlg001/tlg2200.tlg001.opp-grc1.xml
AlisonBabeu: I think sub is actually better, and will change that and try to stick with it in the future.

lcerrato (approved):

Issue #2563, opened 2021-11-29
Textgroup standardization-Removing dates from author names.


AlisonBabeu:
Also will be doing this in CSEL textgroups.

AlisonBabeu:
This work was completed so I am closing this issue.

PR #2564, opened 2021-11-29
Textgroup_name_standardization

AlisonBabeu:
Updating textgroups so they don't include dates
Per issue #2563.

### DIFF
--- a/data/tlg0640/__cts__.xml
+++ b/data/tlg0640/__cts__.xml
@@ -1,3 +1,3 @@
 <ti:textgroup xmlns:ti="http://chs.harvard.edu/xmlns/cts" urn="urn:cts:greekLit:tlg0640">
-  <ti:groupname xml:lang="lat">Alciphron Rhet., Soph.</ti:groupname>
+  <ti:groupname xml:lang="lat">Alciphron</ti:groupname>
 </ti:textgroup>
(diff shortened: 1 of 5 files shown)


data/tlg0640/__cts__.xml
```diff
@@ -1,3 +1,3 @@
 <ti:textgroup xmlns:ti="http://chs.harvard.edu/xmlns/cts" urn="urn:cts:greekLit:tlg0640">
-  <ti:groupname xml:lang="lat">Alciphron Rhet., Soph., AD 2/3</ti:groupname>
+  <ti:groupname xml:lang="lat">Alciphron Rhet., Soph.</ti:groupname>
```
lcerrato, line 2: I wonder if we shouldn't spell out what "Rhet." and "Soph." are?

data/tlg2583/__cts__.xml
```diff
@@ -1,3 +1,3 @@
 <ti:textgroup xmlns:ti="http://chs.harvard.edu/xmlns/cts" urn="urn:cts:greekLit:tlg2583">
-  <ti:groupname xml:lang="eng">Manetho Astrologus, 4th. century CE</ti:groupname>
+  <ti:groupname xml:lang="eng">Manetho Astrologus</ti:groupname>
```
lcerrato, line 2: Probably a Latin attribute?

lcerrato (changes requested):

data/tlg0640/__cts__.xml
AlisonBabeu: Probably. I also just looked this author up and the canonical authority name seems to be Alciphron, I think this is just from the TLG name, I should have changed it in the list.

data/tlg2583/__cts__.xml
AlisonBabeu: I would think so, I will change that.

lcerrato (approved):

PR #2567, opened 2021-12-01
(tlg1271)_all_files

AlisonBabeu:
Updated textgroup.
Updated cts_.xml files and bibliographic descriptions.
Used Clement of Rome for the letter and Clemens Romanus for attributed texts as the author.


lcerrato (approved):

lcerrato:
@AlisonBabeu 
There are more works coming in the workflow so this may need to be revisited. I see we have canonical name differences here and in canonical-greekLit.

PR #2570, opened 2021-12-02
(urn.cts.greekLit.tlg0529.tlg002.1st1K-grc1) new edition

### LINKED ISSUE #2543: (urn.cts.greekLit.tlg0529.tlg002.1st1K-grc1) new edition
[urn.cts.greekLit.tlg0529.tlg002.1st1K-grc1.xml.zip](https://github.com/OpenGreekAndLatin/First1KGreek/files/7349541/urn.cts.greekLit.tlg0529.tlg002.1st1K-grc1.xml.zip)

charhoughton:
fix #2543


lcerrato:
@charhoughton 
Good to see the checks working here! Keep up the submissions.

lcerrato:
@charhoughton 
I see two issues (aside from minor stuff in the header) that I don't think you had any control over:

We need to document the superscript digit thing for future Lace work —these were added with some unusual font encoding (a different Unicode block) which makes it hard to find and replace, so this may be something we really need to watch in Lace in the future. When the volunteers don't use the usual digits, and instead use this weird Unicode block for superscript digits, these characters should technically be encoded as another language or be converted to regular old numbers and marked as superscript in the encoding (preferable). The problem is, that I don't see these in TLG so I don't think we need them at all. (I started to fix these markings in another text but it was way too much labor for something I don't see elsewhere). They don't seem to be important to the body of text and look like annotations.

Also, we can't have the large third section here — it's way too big to be useful. I will add another layer of divs to fix this.

charhoughton:
@lcerrato thanks. The superscript characters are definitely linked to annotations or an app. crit., and I'll make sure to just remove them in the future. I wasn't entirely sure what to do with the last huge div, since it's marked as just one fragment in the text, so thanks for solving that as well.

lcerrato:
did a separate update for the digits in case that ever needs to be rolled back.

lcerrato:
@charhoughton 
Yes, I've seen a couple of texts with the digits. I have a regular expression that just looks for non-greek, which is fine when there is no Latin (and actually spots things like capital letters that are supposed to be Greek but really are English). Otherwise, I have some for these particular script blocks. 
Very strange.

Both of these are downstream from the usual work, so it's not something you have to worry about.

One thing I have noticed is that Lace does not export a top level `editor` tag so I try to add that.

PR #2596, opened 2022-01-27
Update tlg5034.tlg001d.perseus-grc1.xml

PRoryCallahan:
1.) Removed extraneous page number notes.
2.) Removed extra returns/spacing
3.) Made consistent <p> tags to all scholia (a., b., etc.)
4.) Moved MS notations from (presumed) OCR position to where Drachmann intended to be read in text
5.) few <lb/> should have been <p> and vice versa
6.) caught incorrect line numbers
7.) caught a few things missing from transcription


lcerrato:
@PRoryCallahan 
Thanks for this. In order to run checks, this should be submitted via a branch rather than a repo fork, I believe. We can pull it offline for further testing and to view the changes but that will take extra time.

PRoryCallahan:
Thanks, Lisa, for the guidance. I will close this request down and run it again on a branch.

PR #2616, opened 2022-03-15
tlg2036_tlg4021

AlisonBabeu:
Made minor CTS changes and renumbered URNs as they were already assigned in the catalog.


data/tlg2036/tlg001/__cts__.xml
```diff
@@ -1,7 +1,7 @@
-<ti:work xmlns:ti="http://chs.harvard.edu/xmlns/cts" groupUrn="urn:cts:greekLit:tlg2036" xml:lang="grc" urn="urn:cts:greekLit:tlg2036.tlg001">
+<ti:work xmlns:ti="http://chs.harvard .edu/xmlns/cts" groupUrn="urn:cts:greekLit:tlg2036" xml:lang="grc" urn="urn:cts:greekLit:tlg2036.tlg001">
```
lcerrato, line 1: There's an extra space here which is probably the cause of the issues.

lcerrato (approved):

PR #2614, opened 2022-03-05
update tlg5034.tlg001a

PRoryCallahan:
Last one of the Pindar scholia. For my project, I am also going to use the Perseus Pindar (tlg0033) as the base text before making edits/modifications. If I catch anything needing correction, should I be looking at https://github.com/PerseusDL/canonical-greekLit/tree/master/data/tlg0033 for pull requests?

e.g. I see that the title to the Pythian odes are transcribed as πψτηιαν which also displays on Scaife.

Also, super humbling to see change records in those docs going back to the late 80s.

PR #2620, opened 2022-03-30
tlg5034.tlg001a work

lcerrato:
merging new travis.yml into this file set
see also #2614


PRoryCallahan:
@lcerrato This looks great! Thanks so much for all your guidance, and especially for all the work this must have been to re-add the page numbers and then update so much else. I don't have time now, but I will in May to use this as a template to re-add the page numbers and update the other 3 books if you trust me with that.

lcerrato:
@PRoryCallahan 
I've created an issue with (I think) a checklist and other notes: https://github.com/OpenGreekAndLatin/First1KGreek/issues/2626

I may work on these between now and May since it's fresh in my mind. Let's see what happens!

lcerrato:
Note that there were missing opening parentheses in this version which have since been re-added. Just a find/replace error.

PR #2644, opened 2022-05-07
(tlg0062.tlg002-tlg008) Lucian edits

helmadik:
Done some checking on these OCR Lucian files; minor corrections; placed footnotes in line instead of Loeb page bottom. Removed section numbers from running text. In 002 only, removed newlines within paragraphs.


lcerrato:
see #2659

PR #2659, opened 2022-07-22
(tlg0062.tlg002-tlg008) Lucian edits 

lcerrato:
fix #2644 
incorporated prior changes and added
header edits, xml base add, note lang attribute, formatting cleanup


lcerrato:
fix #2644 
changes incorporated here

Issue #2667, opened 2022-09-12
(tlg3135)_tlg3135.tlg004_tlg3135.tlg005


AlisonBabeu:
So while updating the tlg3135.tlg001.opp-grc1 to 1st1K to avoid conflict with the catalog, I "rediscovered" an issue I noted at the time I catalog this work about 5 years ago.  Works tlg 3135.tlg004 and tlg3135.tlg005 are using TLG IDs that have been given to other works. I know we aren't in the practice of being consistent with TLG IDs if something gets misnumbered along the way but  I was wondering if we should create different IDS for these works, especially since they seem to be tightly related to tlg3135.tlg001, 
one is the Argumenta of the work and the other is a Chronologia that accompanies it. I was thinking of renumbering these too
tlg3135.tlg001a and tlg3135.tlg001c in ways we have done before.

Any thoughts @lcerrato?

lcerrato:
@AlisonBabeu 
I'm not following precisely, which is most likely me.
What would these works be if you were starting from scratch?

lcerrato:
@AlisonBabeu 
So I searched on these works and I see that TLG has them as part of 001, so doing 001a and 001b or merging would be better than the current IDs probably.

AlisonBabeu:
Hi @lcerrato I had forgotten to search on them. The TLG is strange with this work since it uses tlg03135.tlg001 for books 1-12 and then tlg3135.tlg002, which is used for books 13-18. I had split them up in the catalog data but had planned on possibly just switching the metadata to use tlg3135.tlg001 since I would rather have the catalog data be consistent with Scaife rather than with the TLG.  Should both of these "Works" just be merged into one file for tlg3135.tlg001? Since I didn't find them in my search, where do tlg3135.tlg004 and tlg3135.tlg005 occur in the TLG edition of tlg313.tlg001?  Thanks for your help!

lcerrato:
@AlisonBabeu 
I didn't look into where they occurred precisely but it was not intuitive. (The Greek also looked a little off from what we have). I would have to revisit the TLG to see what's up.

AlisonBabeu:
hi @lcerrato the pull request didn't work so I'm going to undo it and wait until you have a chance to see where the two works should be integrated. Thanks!

lcerrato:
@AlisonBabeu 
There is a significant difference in the structure of the files we have and the online editions, and even a difference in the structure between 3135.001 (volume/page) and 3135.002 (page).

[...]

PR #2668, opened 2022-09-13
(tlg3135).tlg3135.tlg001

AlisonBabeu:
Re #2667 I am merging tlg3135.tlg004 and tlg3135.tlg005 into tlg3135.tlg001 since the Argumenta and Chronologia have never been considered separate works from the Epitome Historiarum. 

@lcerrato I'm not sure if this merge will work so if this file fails I'll go back and make sure my container units and other items are valid. Thanks!


lcerrato:
@AlisonBabeu 
I think this is moot, but for future reference,—

1) you have two documents with a CTS structure of **chapter/book** and a third with **section**, so that would have to be resolved within the refsDecl and div structure. The chronology can't be one level= section because your refsDecl indicates there are two levels and neither is called section.

2) you have added in the two additional works without creating a new container, which results in the errors of duplicate nodes (passages) as seen in the log report: https://app.travis-ci.com/github/OpenGreekAndLatin/First1KGreek/builds/255517642

So, both the original file and the argumentum have the same structure (book/chapter) and since the argumentum is notations on the original edition, the structure numeration is identical in places. Essentially, you've added a second, duplicate set of passages without a container level superseding these that would differentiate between them.

3) To make this work, you have to a) add a new container level that indicates there are three parts here (=works), and name these; and b) you have to add two levels to the chronology and adjust it so that "section" becomes "chapter", so that you'd have a dummy "book" level (just one book) and then the sections are now chapters. (Or do the alternative with the other two works.)

TL;DR is that adding files to the end of the edition doesn't play nice with the CTS structure because you then do not have unique divs/containers throughout the new file.

lcerrato (approved):

PR #2707, opened 2023-03-22
(tlg0643)_all_files

AlisonBabeu:
Updated CTS metadata-thinking on changing the URNs as we have started to move away from using Work URNs to represent indexes, and sigla.

### DIFF
--- a/data/tlg0643/opp001/__cts__.xml
+++ b/data/tlg0643/opp001/__cts__.xml
@@ -2,7 +2,8 @@
   <ti:title xml:lang="mul">Preface to Anonymi Londinensis ex Aristotelis iatricis Menoniis et aliis medicis eclogae</ti:title>
   <ti:commentary urn="urn:cts:greekLit:tlg0643.opp001.1st1K-lat1" workUrn="urn:cts:greekLit:tlg0643.opp001" xml:lang="lat">
     <ti:label xml:lang="mul">Preface to Anonymi Londinensis ex Aristotelis iatricis Menoniis et aliis medicis eclogae</ti:label>
-    <ti:description xml:lang="mul">Preface to Anonymi Londinensis ex Aristotelis iatricis Menoniis et aliis medicis eclogae, Diels, Reimer, 1893</ti:description>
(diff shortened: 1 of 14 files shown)


data/tlg2016/tlg001/__cts__.xml
```diff
@@ -11,11 +11,11 @@
 			Cambridge University Press. 1891.</ti:description>
 	</ti:edition>
 	
-	<ti:edition urn="urn:cts:greekLit:tlg2016.tlg001.1st1K-lat1" 
+	<ti:translation urn="urn:cts:greekLit:tlg2016.tlg001.1st1K-lat1"
```
lcerrato, line 14: Is this a translation? I thought there were both original Latin and Greek versions of the Passion (not sure the Latin is a translation of the Greek or vice versa).

data/tlg2016/tlg001/tlg2016.tlg001.1st1K-lat1.xml
```diff
@@ -80,7 +80,7 @@ schematypens="http://relaxng.org/ns/structure/1.0"?>
 	
 	<text>
 		<body>
-			<div type="edition" n="urn:cts:greekLit:tlg2016.tlg001.1st1K-lat1" xml:lang="lat">
+			<div type="translation" n="urn:cts:greekLit:tlg2016.tlg001.1st1K-lat1" xml:lang="lat">
```
lcerrato, line 83: same question as above. I think the Latin is original and the Greek is a translation.

data/tlg2016/tlg001/__cts__.xml
lcerrato: I only know because I did a project on it, but this is probably more akin to Bodin, where we have two editions/versions. There was a lost Latin original (we presume).

lcerrato:
@AlisonBabeu 
That is something that should be raised in the software discussion, I think. Aren't there other examples of this?

AlisonBabeu:
@lcerrato Just seeing this again I think there are other examples of this, but haven't tracked them down yet!

lcerrato (approved):

PR #2718, opened 2023-05-04
(tlg0086.tlg003).grc1 and grc2

AlisonBabeu:
Updated cts_xml for two grc files.


lcerrato (approved):
approved pending hook tests

PR #2724, opened 2023-05-24
Tlg0018 grc2

gregorycrane:
The best way to check these files is probably to look at the versions  in https://github.com/gregorycrane/philoalex/tree/main/cohn-corr. That's where the full <corr><sic>ORG</sic><corr>CORRECTION</corr></choice> text is.

There may be problems here but what we have online is terrible. These are certainly much better.


data/tlg0018/tlg001/tlg0018.tlg001.opp-grc1.xml
```diff
@@ -103,966 +104,963 @@
...
 Eus 9 τὰ δ’ ὅσα MFGHR, τὰ δ’ ὅσ’ ABP, ὅσα δὲ Eus κατὰ π. τ. V:
-κατὰ om. ceteri 10 γεγονὼς F οἰκίωσις V, ὁμοίωσις M 10. 11 ἀπεριμάχητον
-FGH: περιμάχητον ceteri 11 κατασκευάζον (ον ex ων corr.) V:
-ἐστὶ MABP, κατασκευάζειν FGH Eus τῶνδε V, τῶ δὲ M 12 ὡς
-τὸν ἔφορον V, sed ὡς rursus deletum ἔφορον] φόβον Eus 13 πάντ’ V Eus:
-τὰ πάντ’ M, τὰ πάντα ceteri γε om. Eus Μωυσῆς VF: Μώσης ceteri
-15 τὰ αὐτὰ codd. ὃν] ὄντα M 15. 16 κατὰ ταῦτα ὄντων μὲν ἀοράτων καὶ
-νοητῶν V 16 συγγενῆ Eus 17 γένεσιν om. Eus καὶ οἰκεῖον Η, τὸ
-οἰκεῖον V εἴτ’ ἐφήμισεν V ἀόρατός V 18 ἀναγκαῖος VR
+κατὰ om. ceteri 10 γεγονὼς F οἰκίωσις V, ὁμοίωσις M 10. 11 ἀπεριμάχητον
```
lcerrato, line 232: There seems to be Latin interspersed here.

gregorycrane: that's the app crit, which has Latin interspersed.

gregorycrane:
I really like the "base" element that I see in the newer files.

I should also tag the damned bibls -- they weren't tagged before but I should add them.

the textual notes are "as is" -- very cool but a black hole that i can't deal with yet.

PR #2725, opened 2023-05-24
first translation added

gregorycrane:
I added the translation for one work of Philo and finally got it to pass Hooktest's dark arts. This passes on my Hooktest and now I will make sure it passes when I push it to a branch on OGL

### DIFF
--- a/data/tlg0018/tlg001/__cts__.xml
+++ b/data/tlg0018/tlg001/__cts__.xml
@@ -4,4 +4,10 @@
     <ti:label xml:lang="lat">De Opificio Mundi</ti:label>
     <ti:description xml:lang="mul">Philo Judaeus. Cohn, Leonard, editor. Opera que supersunt, Volume 1. Berlin: Reimer, 1896.</ti:description>
   </ti:edition>
+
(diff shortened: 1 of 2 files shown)


lcerrato:
@gregorycrane 
Happy to help with cts metadata format and up to date headers. I see this was already pushed, but @AlisonBabeu and I have standard formats for the cts files and the header that we can show you and/or add input on so that we don't have to go back and correct.

We have the title which is the catalog title, the labels are the titles in the  book(s) themselves (so usually Greek for Greek, not Latin or English), etc.
https://github.com/PerseusDL/tei-conversion-tools/wiki/cts-textgroup-and-work-metadata-files

The header is incomplete and not updated (I see things like "la" and the various titles. Whatever was created in 2014 is not going to be all that great.

It would be ideal to be assigned as a reviewer because I could change all of this here in GitHub to show you an example before the push rather than doing this back and forth after the fact which creates conflicts.

Also, we create catalog issues for new editions at the time of push as otherwise the new works are not in the catalog. That's one of the final workflow steps.

I can make edits now but I do not want to affect later pulls on your branch so please let either of us know if we can provide an example file or other input.

Issue #2753, opened 2023-07-22
(tlg0018-yonge) C D Yonge English translations


gregorycrane:
I am building a pull request for the Yonge English translations of Philo (new addition). A separate one for the Berlin Greek updated files will follow.

- \<cRefPattern\>: I now find none without n attribute
- We now use Philo Judaeus (although we should update all our files to the more modern Philo of Alexandria)
- We use the original publisher, Henry G. Bohn (rather than the later George Bell printings)
- each work has a URL that points to the page on which it begins in a Archive.org scan, e.g., tlg0018.tlg030 has https://archive.org/details/worksphilojudaeu04philuoft/page/61/mode/1up. 
- I *could* also put <ref/> tags with links on each page number but the <ref/> tags can't appear in as many places as <pb/> tags can.
- I can't find the controlled vocabulary for <resp/> tags, but I copied <code>\<resp>Digital conversion and editing\</resp>
</code> from a recent file (tlg0550)
- I have moved more detailed info to the \<revisionDesc/>
- Notes have xml:lang attributes.
- I am using 1st1k-eng1 which I got from Alison.
- We now have '\<editor role="translator">' for Yonge.

AlisonBabeu:
hi @gregorycrane so I looked over these two files there. Just two comments on the English file at this point fromme: 

1.) You have the titles mixed up in that you have the title of the volume in the header at line 7 `(<title xml:lang="eng">The Works of Philo Judaeus</title>)`, where you should have the title of the individual work.

This should be in the structured bibl section, where you have `<title xml:lang="eng">On the Creation of the World</title>`, so you just need to reverse those. 

2) You should update the file name from `<idno type="filename">philo_1_1854.xml</idno>`

Two comments for the Greek file:
1)  For the long haul we are updating all of the opp URNs for the Greek files to 1st1K, that way we don't have conflict with metadata in the catalog, but that is something I can do at the end when I add updated catalog metadata for all the translations. 

2) We have also started standardizing the metadata in the <biblStruct> sections so that it isn't Latin anymore so we would update to: `    <publisher>Reimer</publisher>
    <pubPlace>Berlin</pubPlace>` but that is also something I'm happy to do in a later pass.

lcerrato:
Further work needed:

- make sure Greek is declared as language in header
- check ref target
- duplicate notes
- note tags do not need language or n attribute; should have place and resp
- breaks around pb's
- straight apostrophes should be curly
[...]

PR #2757, opened 2023-07-26
(tlg0018.yonge2) -- adding C D Yonge's translations of Philo

gregorycrane:
Comments are in https://github.com/OpenGreekAndLatin/First1KGreek/issues/2753


AlisonBabeu:
So hi @gregorycrane and @lcerrato , two comments to start:

1) I'm not sure why this pull request is not working in terms of the Hooktest, all of the 18 English translations are giving the error:  "Naming conventions". I double checked to make sure the URNs were correct in both the cts_.xml file and the TEI-XML file, and that seems to be correct. And there are other working XML files in 1st1K with the extension 1st1K-eng1

2) Titles are still reversed in all of these files in that all of them have 
'<title xml:lang="eng">The Works of Philo Judaeus</title>' in the main header, instead of the individual work title as listed in the cts_.xml files, and the titles of the works are instead found in the '<biblStruct> ' which is where the volume title should be.

data/tlg0018/tlg001/__cts__.xml
```diff
@@ -1,7 +1,14 @@
 <ti:work xmlns:ti="http://chs.harvard.edu/xmlns/cts" groupUrn="urn:cts:greekLit:tlg0018" xml:lang="grc" urn="urn:cts:greekLit:tlg0018.tlg001">
   <ti:title xml:lang="lat">De Opificio Mundi</ti:title>
-  <ti:edition urn="urn:cts:greekLit:tlg0018.tlg001.opp-grc1" workUrn="urn:cts:greekLit:tlg0018.tlg001" xml:lang="grc">
+  <ti:edition urn="urn:cts:greekLit:tlg0018.tlg001.opp-grc1" workUrn="urn:cts:greekLit:tlg0018.tlg001">
```
lcerrato, line 3: xml:lang is required here. (See issue from Giuseppe on this)

lcerrato (commented):
It looks like the language attribute was incorrectly removed from all metadata here? This is required for post-processing of files and was recently re-added manually by me throughout the collection for Giuseppe.

See https://github.com/OpenGreekAndLatin/First1KGreek/issues/2737

lcerrato:
@AlisonBabeu @gregorycrane 
The files are not passing as the file names have a lower case "k" and the cts metadata has an upper case "K"

See also my note about the lang attribute removal.

data/tlg0018/tlg001/tlg0018.tlg001.1st1k-eng1.xml
```diff
@@ -0,0 +1,1060 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<?xml-model href="https://epidoc.stoa.org/schema/latest/tei-epidoc.rng" schematypens="http://relaxng.org/ns/structure/1.0"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0">
+<teiHeader xml:lang="eng">
+<fileDesc>
+<titleStmt>
+<title xml:lang="eng">The Works of Philo Judaeus</title>
```
lcerrato, line 7: lang not needed here

data/tlg0018/tlg001/tlg0018.tlg001.1st1k-eng1.xml
```diff
@@ -0,0 +1,1060 @@
...
+<sponsor>University of Leipzig</sponsor>
+<funder>European Social Fund Saxony</funder>
+<principal xml:id="GRC">Gregory Crane</principal>
+<respStmt>
+<persName xml:id="Jouve">Jouve</persName>
+<resp>Digital Conversion and edition</resp>
+</respStmt>
+<respStmt>
+<persName>Gregory Crane</persName>
+<resp>Digital conversion and editing</resp>
```
lcerrato, line 17: Digital conversion implies Lace work, but this file says Jouve. So I would recommend simply Digital editing in this case.

data/tlg0018/tlg001/tlg0018.tlg001.1st1k-eng1.xml
```diff
@@ -0,0 +1,1060 @@
...
+<resp>Technical Advisor</resp>
+</respStmt>
+</titleStmt>
+<publicationStmt>
+<authority>University of Leipzig</authority>
+<idno type="filename">tlg0018.tlg001.1st1K-eng1.xml</idno>
+<availability>
+<licence target="https://creativecommons.org/licenses/by-sa/4.0/">Available under a Creative Commons Attribution-ShareAlike 4.0 International License</licence>
+</availability>
+<date>2023</date>
```
lcerrato, line 36: `<date type="release">2023-08-01</date>`

data/tlg0018/tlg001/tlg0018.tlg001.1st1k-eng1.xml
```diff
@@ -0,0 +1,1060 @@
...
+<publisher>University of Leipzig</publisher>
+<pubPlace>Germany</pubPlace>
+</publicationStmt>
+<sourceDesc>
+<listBibl >
+<biblStruct>
+<monogr>
+<author>Philo Judaeus</author>
+<editor role="translator">Charles Duke Yonge</editor>
+<title xml:lang="eng">On the Creation of the World</title>
```
lcerrato, line 46: This is not the edition title, I think

lcerrato (commented):
I think @AlisonBabeu already mentioned some of this

data/tlg0018/tlg001/__cts__.xml
gregorycrane: I have added that to my automatic checks!

data/tlg0018/tlg001/tlg0018.tlg001.1st1k-eng1.xml
gregorycrane: is there any reason not to keep it?

gregorycrane:
> 

The hooktest on my mac passes 1st1k with lower case k - it seems to be case insensitive.

data/tlg0018/tlg001/tlg0018.tlg001.1st1k-eng1.xml
AlisonBabeu: the reason is because the header is labeled "eng" so that attribute is the default assumption for all other subelements, which is why we specify language such as" "lat" "grc" etc.

data/tlg0018/tlg001/tlg0018.tlg001.1st1k-eng1.xml
gregorycrane: done

lcerrato:
@gregorycrane 
Hook test is not case sensitive. This is a frequent error with the CHS volunteer submissions and one I have seen many, many times.

data/tlg0018/tlg001/tlg0018.tlg001.1st1k-eng1.xml
lcerrato: Yes, superfluous encoding.

gregorycrane: It turns out I put this in specially because we are fussy about having lang attributes. Is is that English is the default? Or do we sometimes need to mark it as English?

data/tlg0018/tlg001/tlg0018.tlg001.1st1k-eng1.xml
gregorycrane: Well, it implies OCR work or some sort of conversion. I used tesseract or OCR text from Hathi now.

gregorycrane:
I have reworked the files and will try yet again.

gregorycrane:
Trying hooktest again.  I should have gotten the other fixes as well. The lower case 1st1k surprised me (because my local system acted in a case insensitive way when I copied files).

lcerrato:
@gregorycrane 
Hooktest is never totally going to cover you as we have texts (like the one I am working on) that can pass but not display properly in the reader, or have missing sections, dropped text, etc. 

A single test PR with a single sample/test file before doing a whole batch is always a very helpful approach to narrow down the items in a batch that need attention.

gregorycrane:
Ok -- the hooktest passed here too!

data/tlg0018/tlg001/tlg0018.tlg001.1st1K-eng1.xml
```diff
@@ -0,0 +1,1060 @@
...
+</p>
+</div>
+
+<div type="textpart" subtype="section" xml:base="urn:cts:greekLit:tlg0018.tlg001.1st1K-eng1" n="8">
+<p>But Moses, who had early reached the very summits of philosophy, <note xml:lang="eng" n="2.1">This is in accordance with the description of him in the Bible, where he is represented as being learned in all the wisdom of the Egyptians. </note> and who had learnt from the oracles of God the most numerous and important of the principles of nature, was well aware that it is indispensable that in all existing things there must be an active cause, and a passive subject; and that the active cause is the intellect of the universe, thoroughly unadulterated and thoroughly unmixed, superior to virtue and superior to science, superior even to abstract good or abstract beauty;</p>
```
lcerrato, line 118: - notes are not numbered when they are inline. (Good idea to indicate the placement here).

- plus, this is note 2.1 as is the next one, so the numbering is both not needed and duplicated

- lang attribute not needed.

data/tlg0018/tlg001/tlg0018.tlg001.1st1K-eng1.xml
```diff
@@ -0,0 +1,1060 @@
...
+<pb n="v.1.p.31"/>
+of the third period, the growth of the beard takes place. The fourth period sees him arrive at the fullness of his manly strength. The fifth seven years is the season for marriage. In the sixth period he arrives at the maturity of his understanding. The seventh period is that of the most rapid improvement and growth of both his intellectual and reasoning powers. The eighth is the sum of the perfection of both. In the ninth, his passions assume a mildness and gentleness, from being to a great degree tamed. In the tenth, the desirable end of life comes upon him, while his limbs and organic senses are still unimpaired: for excessive old age is apt to weaken and enfeeble them all.
+</p>
+</div>
+
+<div type="textpart" subtype="section" xml:base="urn:cts:greekLit:tlg0018.tlg001.1st1K-eng1" n="104">
+<p>And Solon, the Athenian lawgiver, described these different ages in the following elegiac verses:—
```
lcerrato, line 654: This looks like a `<quote>` = a quotation, perhaps in block quote?

data/tlg0018/tlg001/tlg0018.tlg001.1st1K-eng1.xml
```diff
@@ -0,0 +1,1060 @@
...
+
+<l>In seven years from th' earliest breath, </l>
```
lcerrato, line 656: curly quote missing

data/tlg0018/tlg001/tlg0018.tlg001.1st1K-eng1.xml
```diff
@@ -0,0 +1,1060 @@
...
+<l>When God has granted ten times seven, </l>
+<l>The aged man prepares for heaven.</l>
+</p>
+</div>
+
+<div type="textpart" subtype="section" xml:base="urn:cts:greekLit:tlg0018.tlg001.1st1K-eng1" n="105">
+<milestone unit="chapter" n="36"/>
+<p>Solon therefore thus computes the life of man by the aforesaid ten periods of seven years. But Hippocrates the physician says that there are seven <note xml:lang="eng" n="31.1">It is hardly necessary to remind the reader of the description of the seven ages of man in Shakespeare. As You Like It, Act 11. sc. 7. </note> ages of man, infancy, childhood, boyhood, youth, manhood, middle age, old age; and that these too, are measured by periods of seven, though not
+<note xml:lang="eng" n="31.1">It is hardly necessary to remind the reader of the description of the seven ages of man in Shakespeare. As You Like It, Act 11. sc. 7. </note>
+<pb n="v.1.p.32"/>
```
lcerrato, line 683: bad open quote marks

lcerrato (commented):
There are a few lingering issues, none of which prevent completion, but these should be flagged for an additional round of refinement.

The CTS needs a title check (Greek is usually in Greek -- may be better in the Greek PR? @AlisonBabeu )

The notes do not need a lang attribute or numbering (which is not unique anyhow). A place attribute (inline) is helpful but not required.

Quotation marks (single and double) should be curly, not straight. `<q>` is preferred to double quotes. It also appears that there may be block quotes in the texts that should be marked as `<quote>` or `<quote rend="block quote">`

lcerrato:
@gregorycrane @AlisonBabeu 
I think this is just about ready. I'd like to fix the quotation marks (straight/curly, `<q>` `<quote>` and the notes, but can either do that now or in another pass.

I think the Greek edition labels should be Greek, but that can go in the other PR. (Very hard to track two massive overlapping batches!)

Welcome further input here.

lcerrato:
@gregorycrane @AlisonBabeu 
I am happy to fix the things I noticed in the body of the files—I think it would be better to do so now than push back (I really prefer that given the current workflow). 
If no one else has current work, I can do that today/tomorrow and then merge.

lcerrato:
Just noticed that Greek is used but not declared.

lcerrato:
@gregorycrane 
I think this is too labor-intensive to handle via GitHub, given there is another overlapping PR active. I just can't manage to do it efficiently this way.

If you can easily automatically add the missing Greek lang declaration to the header of the remaining files (did the first two), I will flag the batch for further work later. (Make sure to pull before pushing)

The biggest thing is the duplicate/errant notes.

lcerrato (approved):

Issue #2588, opened 2022-01-06
Report an issue: urn:cts:greekLit:tlg0527.tlg005.opp-grc2:21.21


agsper:
[tlg0527.tlg005.opp-grc2.xml](https://github.com/OpenGreekAndLatin/First1KGreek/tree/afda8cd/data/tlg0527/tlg005/tlg0527.tlg005.opp-grc2.xml)

Deuteronomy 21:21 assimilates also verses 22-23.  There is a lot of missing text in v.22 and v.23. They should be distributed and filled out something like this :
<div type="textpart" subtype="verse" n="21"><p>καὶ καὶ λιθοβολήσουσιν αὐτὸν οἱ ἄνδρες τῆς <lb n="21"/><lb>πόλεως αὐτοῦ ἐν λίθοις, καὶ ἀποθανεῖται· καὶ ἐξαρεῖς τὸν πονηρὸν<lb>ἐξ ὑμῶν αὐτῶν, καὶ οἱ ἐπίλοιποι ἀκούσαντες φοβηθήσονται.</p></div>
<div type="textpart" subtype="verse" n="22"><p>Ἐὰν δὲ γένηται ἔν τινι ἁμαρτία κρίμα θανάτου, καὶ <lb n="22"/><lb>καὶ ἀποθάνῃ, καὶ κρεμάσητε αὐτὸν ἐπὶ ξύλου·</p></div>
<div type="textpart" subtype="verse" n="23"><p>οὐ κοιμηθήσεται τὸ <lb>σῶμα αὐτοῦ ἐπὶ τοῦ ξύλου, ἀλλὰ ταφῇ θάψετε αὐτὸ ἐν τῇ ἡμέρᾳ<lb>ἐκείνῃ, ὅτι κεκαταραμένος ὑπὸ θεοῦ πᾶς κρεμάμενος ἐπὶ ξύλου· καὶ<lb>οὐ μιανεῖτε τὴν γῆν ἢν κύριος ὁ θεός σου δίδωσίν σοι ἐν κλήρῳ.</p></div></div>

Issue #2585, opened 2022-01-06
Report an issue: urn:cts:greekLit:tlg0527.tlg004.opp-grc2:30.16


agsper:
[tlg0527.tlg004.opp-grc2.xml](https://github.com/OpenGreekAndLatin/First1KGreek/tree/afda8cd/data/tlg0527/tlg004/tlg0527.tlg004.opp-grc2.xml)

Num 30:17 is missing ταῦτα τὰ at the beginning of the verse before δικαιώματα.  (v.17 is assimilated into v.16)

Issue #2583, opened 2022-01-05
Report an issue: urn:cts:greekLit:tlg0527.tlg012.opp-grc2:17.29


agsper:
[tlg0527.tlg012.opp-grc2.xml](https://github.com/OpenGreekAndLatin/First1KGreek/tree/afda8cd/data/tlg0527/tlg012/tlg0527.tlg012.opp-grc2.xml)

σαφφὼθ βο·  should read σαφφὼθ βοῶν

Swete page 648  17:29

Issue #2582, opened 2022-01-05
Report an issue: urn:cts:greekLit:tlg0527.tlg002.opp-grc2:20.13


agsper:
[tlg0527.tlg002.opp-grc2.xml](https://github.com/OpenGreekAndLatin/First1KGreek/tree/afda8cd/data/tlg0527/tlg002/tlg0527.tlg002.opp-grc2.xml)

Exo 20:13  Οὐ<lb n="–16"/> σεῖς should be Οὐ μοιχεύ-<lb n="–16"/> σεῖς

Note: this is 20:14 in English Bibles.

[A more widespread problem (see Exo 20:17) is |· <lb n=| that should be |- <lb n=| particularly with prefixes such as Gen 42:14 κατά· σκοποί  for κατά- <lb n=""> σκοποί representing κατάσκοποί [but this is mostly a visual problem, since this data set does not seem set up for doing word searches for uses of κατάσκοπος]

PR #2779, opened 2024-02-16
(tlg0527.tlg001.opp-grc2.xml) replace Latin characters with Greek

kevinkrahn:
Fixed many instances of Latin characters being used instead of Greek, especially for capital letters which appear identical (e.g. APXH instead of ΑΡΧΗ).


kevinkrahn:
As noted in many issues (#2588, #2587, #2585, #2583, #2582, #2581) tlg0527 is littered with errors. I noticed many other errors while fixing these specific issues in Genesis. I'm not sure if these are worth fixing if these files will be replaced in the future.

lcerrato (approved):

lcerrato:
@kevinkrahn I don't think a replacement is coming in the short term but a new version may be warranted.
Thank you for the changes here.

PR #2784, opened 2024-06-04
tlg0057.tlg066_tlg0057.tlg079

AlisonBabeu:
Next batch of Galen file revisions and page links, this time not on master!

### DIFF
--- a/data/tlg0057/tlg075/tlg0057.tlg075.1st1K-grc1.xml
+++ b/data/tlg0057/tlg075/tlg0057.tlg075.1st1K-grc1.xml
@@ -74,7 +74,8 @@
                             <date>1826</date>
                         </imprint>
                     </monogr>
-                    <ref target="https://catalog.hathitrust.org/Record/009723957">HathiTrust</ref>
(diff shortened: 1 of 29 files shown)


data/tlg0057/tlg075/tlg0057.tlg075.1st1K-grc1.xml
```diff
@@ -74,7 +74,8 @@
                             <date>1826</date>
                         </imprint>
                     </monogr>
-                    <ref target="https://catalog.hathitrust.org/Record/009723957">HathiTrust</ref>
+                    <ref target="https://hdl.handle.net/2027/hvd.hxpp8s?urlappend=%3Bseq=391%3Bownerid=27021597764275152-395">HathiTrust</ref><!-- Volume 11 -->
```
lcerrato, line 77: I use an n attribute to add volume rather than a comment
such as n="vol_11"

AlisonBabeu: Gotcha, I will go back and change that in a final pass. Just one last Galen pull request to send your way after this!

PR #1409, opened 2017-06-07
took english translation from hebrew and made new working group heb00…

ahanhardt:
…01. Added appropriate cts files with structuredmetadata. Edited cts file for tlg0527.tlg048 accordingly
Encountered a problem earlier where this issue branch somehow became merged to merge request #1339. So we deleted the branch 133, recreated it, and are requesting a new pull request with the needed changes.

PR #2789, opened 2024-07-10
heb0001.heb010 updating cts_xml

AlisonBabeu:
Updated the CTS_XML file for this translation of Isaias to agree with other works taken from the same printed edition.

I am also contemplating why this record was created with the ID heb0001.heb010 when the English translation of this work from the Greek (in the same book) was given the ID tlg0527.tlg048.opp-eng2. There is some discussion here (https://github.com/OpenGreekAndLatin/First1KGreek/pull/1409).  I'm wondering if they wanted to specify this work had been translated from the Hebrew rather than the Greek, but then of course, this URN means it doesn't align with any of the other versions or translations of Isaias. Thoughts @lcerrato? @gregorycrane ?

### DIFF
--- a/data/heb0001/heb010/__cts__.xml
+++ b/data/heb0001/heb010/__cts__.xml
@@ -5,8 +5,8 @@
     <dct:hasVersion>urn:cts:greekLit:tlg0527.tlg048</dct:hasVersion>
     </cpt:structured-metadata>
   <ti:translation xml:lang= "eng" urn="urn:cts:hebrewlit:heb0001.heb010.1st1K-eng1" workUrn="urn:cts:hebrewlit:heb0001.heb010">
-    <ti:label xml:lang="lat">Isaias</ti:label>
-    <ti:description xml:lang="mul">Septuaginta. The Book of Isaiah According to the Septuagint (Codex Alexandrinus). Ottley, Richard, Rusden, editor. 
+    <ti:label xml:lang="lat">Isaiah (from the Hebrew)</ti:label>
+    <ti:description xml:lang="mul">The Book of Isaiah According to the Septuagint (Codex Alexandrinus). Ottley, Richard, Rusden, editor. 
       Cambridge: C.J. Clay and Sons, 1904.</ti:description>
   </ti:translation>
 </ti:work>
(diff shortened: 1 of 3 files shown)


lcerrato:
@AlisonBabeu I think there was some discussion of a special `<xref>` in the header or elsewhere to indicate the source for the translation? 

There already is a Hebrew text in the SV which is not cross referenced here. I think the issue is that tlg0527 is specifically the Septuagint not the Hebrew, right? And we have Hebrew that currently has no relationship to those texts.

Wouldn't this make the description wrong? This implies it is a translation of the Septuagint.
`Septuaginta. The Book of Isaiah According to the Septuagint (Codex Alexandrinus).`

data/heb0001/heb010/__cts__.xml
```diff
@@ -6,6 +6,7 @@
     </cpt:structured-metadata>
   <ti:translation xml:lang= "eng" urn="urn:cts:hebrewlit:heb0001.heb010.1st1K-eng1" workUrn="urn:cts:hebrewlit:heb0001.heb010">
     <ti:label xml:lang="lat">Isaias</ti:label>
```
lcerrato, line 8: I think the label would be `Isiah (from the Hebrew)` as that is the work title in print.

data/heb0001/heb010/__cts__.xml
```diff
@@ -6,6 +6,7 @@
...
-    <ti:description xml:lang="mul">Hebrew Bible, Isaias, Ottley, Cambridge, 1904</ti:description>
+    <ti:description xml:lang="mul">Septuaginta. The Book of Isaiah According to the Septuagint (Codex Alexandrinus). Ottley, Richard, Rusden, editor.
```
lcerrato, line 9: I would not include Septuginta here as this is apart from that.

data/heb0001/heb010/__cts__.xml
AlisonBabeu: I will change that!

data/heb0001/heb010/__cts__.xml
AlisonBabeu: Will change this.

AlisonBabeu: And you are right @lcerrato there was additional metadata in the cts_.xml file. ` <cpt:structured-metadata>
    <dct:hasVersion>urn:cts:greekLit:tlg0527.tlg048</dct:hasVersion>
    </cpt:structured-metadata>` I guess we will just leave this as is for now, it's just without the matching URNs there is no way to align the texts and I could see someone wanting to align the two English translations.

lcerrato: @AlisonBabeu I'm not opposed to renaming, I'm just wondering how it would work since we can't currently align with either of these: https://scaife.perseus.org/library/urn:cts:ancJewLit:hebBible.isaiah/

lcerrato (approved):

PR #2792, opened 2024-07-18
(new_submissions) adding tlg0544.tlg001.1st1K-eng1.xml file

### LINKED ISSUE #2791: (urn:cts:greekLit:tlg0544.tlg001) Sextus Empiricus translation ingestion
adding contribution from @WDRoush

lcerrato:
fix  #2791


lcerrato:
@AlisonBabeu Would you double check the metadata and header?

AlisonBabeu (commented):
hi @lcerrato the only thing I think that needs changing is the `  <title>Outlines of Pyrrhonism</title>` in the '<biblStruct>' should instead be the title of the published volume itself, namely Sextus Empiricus.

lcerrato:
> hi @lcerrato the only thing I think that needs changing is the ` <title>Outlines of Pyrrhonism</title>` in the '' should instead be the title of the published volume itself, namely Sextus Empiricus.

I checked and we had examples of both, where this is the Volume subtitle.  We can/should revise as needed.

AlisonBabeu (approved):
I don't see any other things in the metadata that need changing!

PR #2812, opened 2024-12-08
Update tlg0026.tlg004.1st1K-grc1.xml

mgbilby:
hyphenated word at page break between p288 and 289 in Blass, Aeschinis Orationes; led to erroneous word token "fam" in recent OGA edition


mgbilby:
Awesome. Thank you!

PR #1945, opened 2017-12-07
Adding tlg1181.tlg001 from Schiefsky's file to FirstIK-grc

rebeccadraughon:
Changed tlg1181.tlg001 from Schiefsky's list to DDD compliance; added to First1K-grc.

PR #2822, opened 2024-12-18
(tlg0087.tlg035.1st1K-grc1.xml) orphaned legacy file work

lcerrato:
integrating changes from old PR #1945


AlisonBabeu (approved):
HI @lcerrato this looks good!

lcerrato:
@AlisonBabeu  I realize I should be looking at the pick list for these odds and ends I found.

PR #2823, opened 2024-12-18
(text_corrections) regularizing data

lcerrato:
regularizing this von Spengel RG edition data presentation


lcerrato:
@AlisonBabeu I discovered some irregularities when working on that last orphaned file.
I don't know if this is the best representation, but I made it consistent at least.

AlisonBabeu (approved):
hi @lcerrato this looks solid, thanks for fixing up. I'm going to go fix up another author whose cts metadata from these files needs some cleanup.

PR #2838, opened 2025-06-10
tlg0658.tlg001.perseus-grc1.xml replace question marks with greek

IsaacBennettSmith:
I've gone through the '??' parts and checked the source text for the correct greek


lcerrato (approved):
@IsaacBennettSmith 
Thank you for all of this work!

PR #2849, opened 2025-10-22
Fix various typographical errors in XML file

mgbilby:
pages 32-33


lcerrato:
@mgbilby Are the `<` and `>` additions in this work? If so I will convert to the proper tags.

PR #2858, opened 2026-03-25
Removed consecutive spaces occurring in some date tags.

chrisdrymon:
Removed consecutive spaces occurring in some date tags.


lcerrato:
@chrisdrymon Thank you for this.

I don't know about the file capture, but dates should preferably be `<date when="...">....</date>` 

Further, it looks like in this text these are marginal notes rather than just dates:

`<note resp="editor" place="margin"><date when="0324">A. D. 324.</date></note>`

I can make these changes to the file shortly.

lcerrato (approved):

PR #2863, opened 2026-05-25
Fix ἐπ̓ and ὑπ̓ elisions

chrisdrymon:
Fixes ἐπ̓ and ὑπ̓ elisions where the π was taking a breath mark instead of a proper elision mark occurring. They are now ἐπ’ and ὑπ’. Only occurs across 17 files.


lcerrato (approved):
@chrisdrymon 
Thank you!

PR #2865, opened 2026-05-25
Fix ἀλλ̓ elisions

chrisdrymon:
Replace smooth breathing mark (U+0313) on λ with apostrophe (U+2019) in elided forms of ἀλλά.


lcerrato:
@chrisdrymon Is it worth re-running this to insert spaces where missing, with a check for things that break on a line. 
I see a few pre-existing spacing problems in just the first text.

lcerrato:
fix applied via #2881

PR #2868, opened 2026-05-25
Fix μετ̓ elisions

chrisdrymon:
Replace smooth breathing mark (U+0313) on τ with apostrophe (U+2019) in elided forms of μετά.


lcerrato:
fixed via #2884

PR #2869, opened 2026-05-25
Fix παῤ elisions

chrisdrymon:
Replace smooth breathing mark (U+0313) on ρ with apostrophe (U+2019) in elided forms of παρά. Skips cases where the breathing on ρ marks the older double-rho orthographic convention (παῤῥ-) rather than elision.


lcerrato:
fixed via #2886

PR #2873, opened 2026-05-25
Fix ἵν̓ elisions

chrisdrymon:
Replace smooth breathing mark (U+0313) on ν with apostrophe (U+2019) in elided forms of ἵνα.


lcerrato:
fixed via #2886

PR #2875, opened 2026-05-25
Mark Hultsch excursus chapter as xml:lang="deu"

chrisdrymon:
Chapter 14 is F. Hultsch's German editorial excursuses, not source Greek. Mark it as German rather than Greek. The prose is German throughout with occasional Greek lines quoted from the source interspersed.


lcerrato (commented):
In the absence of precise encoding of included Greek, it would  likely be "mul" rather than "deu" in this case perhaps.

